PR #4, opened 2014-11-02
added inline monospace and colorize source block

vincentvdk:
I added some highlighting that I currently use in vim.

### DIFF
--- a/grammars/asciidoc.cson
+++ b/grammars/asciidoc.cson
@@ -192,6 +192,17 @@
         'match': '`[^`]+`'
         'name': 'markup.raw.inline.asciidoc'
       }
+
+      # Matches quoted text Monospaced
+      #
+      # Examples
+      #
+      #    This is +monospaced+ text.
+      {
+        'match': '(?<=^|\\s)(\\+[^\\+]+\\+)'
+        'captures':
+          '0': 'name': 'markup.raw.inline.asciidoc'
+      }
     ]
   'codeBlocks':
     'patterns': [
--- a/stylesheets/asciidoc.less
+++ b/stylesheets/asciidoc.less
@@ -35,5 +35,16 @@
     .table.spec {
       color: @syntax-text-color-unobtrusive;
     }
+
+    .markup.raw.inline {
+      font-weight: bold;
+      font-family: monospace;
+      color: @syntax-text-color-unobtrusive;
+    }
+
+    .support {
+      color: orange;
+    }
+
   }
 }


gallir:
@vincentvdk Are you going to merge and publish this request? I was going to do the same.

vincentvdk:
@gallir I can't merge since i've no permissions to do so..

mojavelinux:
Looks like it might need a rebase. @antonmoiseev, do you need @vincentvdk to rebase?

antonmoiseev:
I can rebase, however I would really like to see review comments to be addressed before merging it.

JFYI, we just started to work on a new book at Farata. As you may know we used asciidoctor before for the [book](http://enterprisewebbook.com/) and for a bunch of [training materials](http://www.eventbrite.com/o/farata-systems-37982026#past_events), so we are happy the publisher allowed us to use asciidoc for the new book as well.

I use Atom as my primary editor and see how much it lacks the decent asciidoc syntax support at the moment. I'm really looking for a chance to make some progress here in next couple of months.

mojavelinux:
> we are happy the publisher allowed us to use asciidoc for the new book as well.

Fantastic!

>  I'm really looking for a chance to make some progress here in next couple of months.

:beers:

nicorikken:
This seems like a well thought out, well discussed feature.
The styling discussion in particular relates to other features. Maybe such conclusions can be included in a HACKING.adoc file or something.

Are there any blockers for this merge other than incompatible branches?

mojavelinux:
> The styling discussion in particular relates to other features. Maybe such conclusions can be included in a HACKING.adoc file or something.

:+1: 

> Are there any blockers for this merge other than incompatible branches?

We shouldn't be highlighting monospace using + symbols anymore as that is legacy AsciiDoc. If it's possible, we'd only enable this if `:compat-mode:` is set. However, I'm not sure if the grammar can be changed based on settings in the document. Therefore, I don't think we should do it.

Having said that, I am okay with highlighting + symbols as passthrough. In that case, we just need to name the match differently (though I'm sure I understand the name "markup.raw.inline.asciidoc". Maybe we need some better names here.

nicorikken:
Then we need some proper tests, to ensure the passthrough doesn't break.

nicorikken:
Having given this a second thought, highlighting as passthrough might be a valid option, although it might lead to 'interesting' conflicts between regex's, which are hard to even notice. In that sense I guess its best to leave out this backwards compatibility (compat-mode) similar to the approach taken for file extensions. Therefore I believe this PR can be closed. What do you think?

mojavelinux:
My position is that we don't want to encourage the use of compat mode. Now that the modern syntax is supported on GitHub (and generally everywhere else), I think we just look forward.

I do think it's worth highlighting the plus version of the passthrough. We may want to persue that as a separate issue, though, so the subject is clear.

nicorikken:
I just created PR #85 to clarify and document this decision, as we both agree.

nicorikken:
The lack of support for `compat-mode` has now been documented, and therefore this feature will not be supported any time soon. All contributors, thanks for all contributions in code and in discussion, it did bring value to the project. I'll close this PR now.

mojavelinux:
:+1: 

@vincentvdk and others, please let us know how the new syntax support works for you. If you have any troubles, don't hesitate to open new issues.

Issue #11, opened 2015-06-09
Snippets - Periods get in the way


adam820:
Snippets are tending to get in the way of regular writing. When ending a sentence, it's wanting to interpret the period as the start of a literal block. When hitting enter to drop to a new line, it pops in a literal block:
![image](https://cloud.githubusercontent.com/assets/1794794/8067988/2c4df9c8-0ebf-11e5-969c-f7e81cdc4478.png)
![image](https://cloud.githubusercontent.com/assets/1794794/8068031/5ef440b2-0ebf-11e5-959b-e687cdc19c0b.png)

This can get aggravating while trying to do bulleted lists, etc:
![image](https://cloud.githubusercontent.com/assets/1794794/8068057/88a22a82-0ebf-11e5-8f1e-47c090eb0638.png)

A workaround is to either hit Esc after each period, or otherwise disable the snippets package, but that doesn't seem user-friendly to have to do in the long-run.

Would it be possible to detect only a single period as not being the prefix, or otherwise changing the prefixes to something less ambiguous? Perhaps something like 'litbl' for Literal Block, 'exbl' for Example Block, etc.

mojavelinux:
Is there anyway to indicate that a snippet should only trigger at the start of the line? Because the only way a literal block delimiter makes any sense is if it's at the start of a line.

mojavelinux:
Totally agreed, btw.

adam820:
Changing the syntax to start with "bl" might help the usability, since all the options would pop into the autocomplete:

![image](https://cloud.githubusercontent.com/assets/1794794/8068772/0da1b1d6-0ec4-11e5-9933-4d16eeac2478.png)

It's also probable that you're unlikely to ever type "blqte" ambiguously.

adam820:
I whipped a pull request if wanted that makes the above noted changes, with a few changes. It boils down to "bl" + the first three letters of the type of delimited block:

`Comment Block: // -> blcom`
`Example Block: == -> blexa`
`Listing Block: -- -> bllis`
`Literal Block: .. -> bllit`
`Passth. Block: ++ -> blpas`
`  Quote Block: __ -> blquo`
`Sidebar Block: ** -> blsid`

If you would like to head down the road of trying to fix the overall issue of the trigger happening on EOL, then reject away.

mojavelinux:
[...]

PR #13, opened 2015-06-15
Change snippet trigger syntax

adam820:
Changing the trigger syntax for the delimited blocks snippets, to avoid things like periods, asterisks, etc. from accidentally triggering a block when attempting to line wrap.

As discussed in https://github.com/asciidoctor/atom-language-asciidoc/issues/11.


antonmoiseev:
Please disregard the failed message. Since we made the last change part of the Atom's API we use in the tests become deprecated. Just pushed the fix.

PR #17, opened 2015-07-07
Formatted block support, blockid and reference elements

neontapir:
This pull request contains support for several additional elements. While CSON doesn't allow functions, the test suite does, so there was some refactoring there as well. If you'd like these in smaller chunks, please let me know.


neontapir:
Also, if you'd like me to squash all those commits, I'm happy to do that too. I meant to do so before I submitted.

antonmoiseev:
Thank you! I'll take a look on the weekend.

grammars/asciidoc.cson
```diff
@@ -30,7 +30,7 @@
   #
   #   === Level 2 Section
   {
-    'begin': '^={1,6}\\s+'
+    'begin': '^={1,6}\\p{Blank}+'
```
antonmoiseev: @neontapir I didn't know about this property, how is it better than `\s`?

grammars/asciidoc.cson
```diff
@@ -64,17 +80,123 @@
...
+  #
+  # Examples
+  #
+  #   ====
+  #   A multi-line example.
+  #
+  #   Notice it's a delimited block.
+  #   ====
+  {
+    'begin': '^(={4,})+?$'
```
antonmoiseev: I've noticed this pattern differ from other block patterns. Why do you need `+?` after the group?

grammars/asciidoc.cson
```diff
@@ -64,17 +80,123 @@
...
+  #
+  # ...................................
+  # Daleks EXTERMINATE in monospace!
+  # ...................................
+  {
+    'begin': '(^\\.{4,}$)'
+    'end': '\\1'
+    'name': 'markup.block.literal.asciidoc'
+    'patterns': [
+      { 'include': [ '#inline', '#codeBlocks'] }
```
antonmoiseev: According to the [manual](http://asciidoctor.org/docs/user-manual/#built-in-blocks-summary) substitution rules do not apply to literal blocks. I think we don't need to include inline and code block patterns here.

grammars/asciidoc.cson
neontapir: \p{Blank} matches a space or a tab

grammars/asciidoc.cson
```diff
@@ -64,17 +80,123 @@
...
+  #
+  # +++++++++++++++++++++++++++++++++++
+  # <s>Could be struck through</s>
+  # +++++++++++++++++++++++++++++++++++
+  {
+    'begin': '(^\\+{4,}$)'
+    'end': '\\1'
+    'name': 'markup.block.passthrough.asciidoc'
+    'patterns': [
+      { 'include': '#inline' }
```
antonmoiseev: I think we don't need includes for passthrough blocks as well as for literal block.

grammars/asciidoc.cson
neontapir: That's fair. I'll test both of the comments in this file against the book I'm writing (my test case) and report back either with a revised pull request or an illustration.

grammars/asciidoc.cson
antonmoiseev: `\s` matches both spaces and tabs as well. However `\s` also matches new lines, but I don't see what issues it may cause.

Anyway it was mostly out of curiosity I'm fine with `\p{Blank}`.

neontapir: As you might suspect, I had a trailing tab on a line and realized it can be a common issue, since most text editors don't display tabs except as blank space.

grammars/asciidoc.cson
```diff
@@ -64,17 +80,123 @@
...
+  # ____________________________________________________________________
+  # D'oh!
+  # ____________________________________________________________________
+  #
+  # [quote, Erwin Schrödinger, Sorry]
+  # ____________________________________________________________________
+  # I don't like it, and I'm sorry I ever had anything to do with it.
+  # ____________________________________________________________________
+  {
+    'begin': '^\\[(quote|verse)(, ([\\w ]+))?(, ([\\w ]+))?'
```
antonmoiseev: I'm not sure whether we should make a separate pattern for each block name. I'd rather keep it simple and have a single one that matches all possible block names. I just don't see much benefit if they are highlighted differently. @neontapir what do you think?

grammars/asciidoc.cson
```diff
@@ -84,6 +206,32 @@
...
   }
 
+  # Matches explicit paragraph styles
+  #
+  # Examples
+  #
+  #   [normal]
+  #   This is a normal paragraph.
+  {
+    'match': '^\\[((normal|literal|listing|TIP|NOTE|IMPORTANT|WARNING|CAUTION|partintro|comment|example|sidebar|source|music|latex|graphviz))\\]$'
```
antonmoiseev: If you agree with [this comment](https://github.com/asciidoctor/atom-language-asciidoc/pull/17/files#r36923465), we could generalize this pattern to match any text within the square brackets.

grammars/asciidoc.cson
antonmoiseev: Sorry but I don't get the issue. A trailing tab shouldn't break the syntax highlighting of a title.

grammars/asciidoc.cson
```diff
@@ -84,6 +206,32 @@
...
+  }
+
+  # Matches section templates
+  #
+  # Examples
+  #
+  #   [appendix]
+  #   This is an appendix.
+  {
+    'match': '^\\[((sect[1-4]|abstract|preface|colophon|dedication|glossary|bibliography|synopsis|appendix|index))\\]$'
```
antonmoiseev: Again if you agree with the comments above we could combine block name patterns into a single one.

grammars/asciidoc.cson
neontapir: I agree, hence \p{Blank}. Let's say you had a line `== Title\t` in your file. The original regex wouldn't match, whereas the \p{Blank} would.

grammars/asciidoc.cson
neontapir: I see your point, we could remove this. The style I submitted only cares about markup.quote. I chose to submit it because without those lines, we're ignoring information that other style writers could potentially use. It's your call, just give the word and I'll remove it.

grammars/asciidoc.cson
neontapir: I thought about that, too. I chose to match just these items because they are the only ones in the specification. By identifying these specifically, the highlighting could help authors detect typos.

grammars/asciidoc.cson
neontapir: I agree, it can be removed.

grammars/asciidoc.cson
```diff
@@ -179,20 +323,108 @@
...
+      #
+      # NOTE: As mentioned in http://stackoverflow.com/questions/1141848/regex-to-match-url,
+      # this regex is not foolproof.
+      #
+      # Examples
+      #
+      #   http://www.docbook.org
+      {
+        'match': '(?<=^|\\s)(\\S+://\\S+\\.(com|net|org|edu|gov)(\/\\S+)?)'
+        'name': 'markup.url.asciidoc'
```
ldez: I think it's better `markup.underline.link.asciidoc`

antonmoiseev: Agree. Fixed after merging PR.

PR #29, opened 2015-08-29
feat: support markdown Headings syntax

ldez:
support markdown Headings syntax

### DIFF
--- a/grammars/asciidoc.cson
+++ b/grammars/asciidoc.cson
@@ -30,7 +30,7 @@
   #
   #   === Level 2 Section
   {
-    'begin': '^={1,6}\\p{Blank}+'
+    'begin': '^((?:=|#){1,6})([ \\t]+)(?=\\S+)'
     'end': '$'
     'name': 'markup.heading.asciidoc'
     'patterns': [
--- a/spec/asciidoc-spec.coffee
+++ b/spec/asciidoc-spec.coffee
@@ -104,6 +104,31 @@ describe "AsciiDoc grammar", ->
     expect(tokens[1]).toEqual value: "An example example", scopes: ["source.asciidoc", "markup.heading.blocktitle.asciidoc"]
     expect(tokens[3]).toEqual value: "=========", scopes: ["source.asciidoc", "markup.block.example.asciidoc"]
 
+  it "tokenizes Mardown-style headings", ->
+    {tokens} = grammar.tokenizeLine("# Heading 0")
+    expect(tokens[0]).toEqual value: "# ", scopes: ["source.asciidoc", "markup.heading.asciidoc"]
+    expect(tokens[1]).toEqual value: "Heading 0", scopes: ["source.asciidoc", "markup.heading.asciidoc"]
+
+    {tokens} = grammar.tokenizeLine("## Heading 1")
+    expect(tokens[0]).toEqual value: "## ", scopes: ["source.asciidoc", "markup.heading.asciidoc"]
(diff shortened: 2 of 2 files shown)


nicorikken:
Nice addition. :+1: 
Do you have a particular reason for choosing `\p{Blank}` over `/s`? As I'm still learning about regexes, I'd like to know if it would make a difference.
Regarding the rest of the commit, I'd like to see the Markdown matching in a separate set of matches, rather than including it inside the normal Asciidoc headers. Or does that result in matching conflicts or performance issues? I guess it's a matter of preference.

mojavelinux:
> Do you have a particular reason for choosing `\p{Blank}` over `\s`?

`\p{Blank}` is another way of writing `[ \t]`, whereas `\s` is a whole variety of whitespace including endlines. Therefore, we should use the former. However, I would prefer if we write it as `[ \t]` to make it more clear (though I think we have to use two backslashes in this context, so `[ \\t]`).

I find the following page extremely helpful for understanding regexp character classes: http://ruby-doc.org/core-2.3.0/Regexp.html

mojavelinux:
I also didn't realize that the `\p{}` references were allowed in the grammar file. That's not typical of JavaScript. I can't find any information about which regex features are permitted in this grammar file. I guess we can derive that information via tests.

ldez:
POSIX expression are not supported by JavaScript.
I've replaced by `[ \\t]` and rebase.

mojavelinux:
:+1:

grammars/asciidoc.cson
```diff
@@ -30,7 +30,7 @@
   #
   #   === Level 2 Section
   {
-    'begin': '^={1,6}\\p{Blank}+'
+    'begin': '^(={1,6}|[#]{1,6})([ \\t]+)'
```
mojavelinux: This can be written more concisely as:

```
'^((?:=|#){1,6})([ \\t]+)'
```

We may even one to do a lookahead for at least one non-space character:

```
'^((?:=|#){1,6})([ \\t]+)(?=\S)'
```

ldez: I take the second expression :wink:

ldez: The second expression breaks the tests.
I've reverted to the first proposition.

mojavelinux:
Looks good to me! @nicorikken, feel free to do the honors of merging.

mojavelinux:
Thanks for updating the PR so quickly @ldez!

mojavelinux:
> The second expression breaks the tests. I've reverted to the first proposition.

I was missing the double backslash required because we're inside a string. The following works:

```
'begin': '^((?:=|#){1,6})([ \\t]+)(?=\\S+)'
```

mojavelinux:
Can you update one more time? Then we'll be ready for merge.

ldez:
It's done!

mojavelinux:
Thank you! Time for merging @nicorikken!

nicorikken:
Merge is done, thanks @ldez and @mojavelinux :+1:

mojavelinux:
:beers: 

Congratulations on your first merge, @nicorikken! I'm excited for the future of this package.

mojavelinux:
And, of course, thank you @ldez for submitting and updating this pull request. Nice teamwork!

nicorikken:
@mojavelinux If I understand correctly, boosting the version number of the package (`package.json`) would allow the new features to become available in the editors of the users. I can handle more of the outstanding pull-requests in the upcoming days, aiming for a new release say next week.

mojavelinux:
Yes!

Issue #26, opened 2015-08-29
Asciidoctor syntax highlighting on lists


Mogztter:
I can see in the code that lists are supported (https://github.com/asciidoctor/atom-language-asciidoc/blob/master/grammars/asciidoc.cson#L141) but the default color is too subtle:

![syntax-highlighting](https://cloud.githubusercontent.com/assets/333276/9561764/a4ed7ce0-4e53-11e5-8890-ee69fa0ccdcd.png)

Also numbered lists are currently not supported ?

ldez:
I have create a PR for bullet color : #30

ldez:
This issue can be close, version 1.0.0 fix the problem.

Mogztter:
![great](https://cloud.githubusercontent.com/assets/333276/14936108/27c69408-0ee8-11e6-8126-bc770330652e.png)
:+1:

PR #30, opened 2015-08-29
LESS optimize syntax

ldez:
- use `&` instead of repeat `.markup`
- fix font smoothing
- remove custom bullet style to use default style for better readability

### DIFF
--- a/styles/asciidoc.atom-text-editor.less
+++ b/styles/asciidoc.atom-text-editor.less
@@ -5,55 +5,63 @@
 atom-text-editor, :host {
   .asciidoc {
 
-    .markup.bold {
-      font-weight: bold;
-    }
+    .markup{
 
-    .markup.italic {
-      font-style: italic;
-    }
-
-    .markup.url {
-      color: mix(red, @syntax-text-color, 20%);
-    }
-
-    .markup.htmlentity {
-      font-style: italic;
-      color: mix(red, @syntax-text-color, 20%);
-    }
-
-    .markup.biblioref,
-    .markup.blockid {
-      font-weight: bold;
-      color: mix(green, @syntax-text-color, 20%);
-    }
-
-    .markup.reference {
-      font-style: italic;
-      color: mix(green, @syntax-text-color, 20%);
-    }
-
-    .markup.admonition,
-    .markup.explicit,
-    .markup.heading,
-    .markup.section {
-      font-weight: bold;
-
-      // Lighten headers for dark themes.
-      & when (lightness(@syntax-background-color) < 50%){
-        color: lighten(@syntax-text-color, 20%);
+      &.bold {
+        font-weight: bold;
       }
 
-      // Darken headers for light themes.
-      & when (lightness(@syntax-background-color) > 50%){
-        color: darken(@syntax-text-color, 20%);
+      &.italic {
+        font-style: italic;
+      }
+
+      &.url {
+        color: mix(red, @syntax-text-color, 20%);
+      }
+
+      &.htmlentity {
+        font-style: italic;
+        color: mix(red, @syntax-text-color, 20%);
+      }
+
+      &.biblioref,
+      &.blockid {
+        font-weight: bold;
+        color: mix(green, @syntax-text-color, 20%);
+      }
+
+      &.reference {
+        font-style: italic;
+        color: mix(green, @syntax-text-color, 20%);
+      }
+
+      &.admonition,
+      &.explicit,
+      &.heading,
+      &.section {
+        font-weight: bold;
+
+        // Lighten headers for dark themes.
+        & when (lightness(@syntax-background-color) < 50%){
+          color: lighten(@syntax-text-color, 20%);
+        }
+
+        // Darken headers for light themes.
+        & when (lightness(@syntax-background-color) > 50%){
+          color: darken(@syntax-text-color, 20%);
+        }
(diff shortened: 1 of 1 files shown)


nicorikken:
@ldez Font smoothing now seems to be handled on an editor-level (which seems to make more sense): https://github.com/atom/atom/issues/1650 So I guess the rendering feature can be removed, and the code should be updated to the latest branch before the merge. Adhering to upstream stylesheets certainly seems to make more sense, especially if custom styles aren't regarded as improvements (as mentioned in issue #26 )
Can some of the project maintainers comment on this more strategic decision regarding styling?

mojavelinux:
Font smoothing and legibility settings always seem to bite us since browsers can't seem to agree on what the defaults are. I'm fine with letting the browser/environment decided on the font-smoothing at least.

Changes to the stylesheet right now are incremental since we still need to resolve https://github.com/asciidoctor/asciidoctor-stylesheet-factory/issues/18 before we have the right level of control. It shouldn't be necessary for the Atom plugin to be undoing styles that it doesn't want. Instead, we should be able to generate exactly the styles we need. But again, there are steps to getting there.

mojavelinux:
I'm fine with this change, though it will need to be rebased to be merged.

ldez:
I've rebase.
Thanks for your review.

mojavelinux:
There seem to be several styles that are being removed. Is this intentional?

ldez:
It's a mistake, sorry.
Too many conflicts...

I've correct errors.

mojavelinux:
Thanks @ldez!

@nicorikken can you confirm these changes look good to you?

nicorikken:
I've reviewed the changes, and in general they look good. I'm however a bit weary of dropping any form of styling for `.markup.list.bullet`, as the highlighting shows the user that the syntax is recognized as valid. I'm not familiar with styling yet. For example I wasn't able to reverse-engineer the styling used in the github flavoured markdown highlighting for lists. @ldez maybe you have another styling proposal for list bullets rather than dropping any highlighting?

ldez:
No color it's a color :wink: 

asciidoctor-nocolor

But I can do this:

asciidoctor-red

nicorikken:
Looks good to me. Similar to the bold styling but without the boldness I assume?

ldez:
It's between 'url' color and 'bold' color.

mojavelinux:
This is good, but at some point we're going to need to look at all these color choices and make sure that they work well with all themes. Just mixing colors against the theme colors doesn't always lead to the best result.

In fact, we probably shouldn't be wiring colors directly into the language. This seems to be something that is done in an editor theme. However, I'm not sure if it's possible to apply multiple themes, so if the provided themes don't support AsciiDoc, then we are back to having to add our own styling.

The best approach is probably to merge these theme changes now, continue to refine our naming conventions for syntax, then eventually get theme authors to incorporate styling for AsciiDoc. So it's an incremental process.

styles/asciidoc.atom-text-editor.less
```diff
@@ -5,55 +5,64 @@
 atom-text-editor, :host {
   .asciidoc {
 
-    .markup.bold {
-      font-weight: bold;
-    }
+    .markup{
+      -webkit-font-smoothing: auto;
```
mojavelinux: I think we should remove this line. This is something that should be provided by the base editor theme.

ldez:
LESS have a very powerful syntax and Atom provide theme variable like `@syntax-text-color`.

All colors are relative to the editor theme variables and the editor compute then on launch.
```less
@import 'syntax-variables';
```

Before my proposal I've test with some others theme like white theme.

mojavelinux:
To your credit, I do appreciate that the colors you've selected are tinted from the theme colors. I still think we'd need to do an overall review once we get all the syntax accounted for to make sure that we following a consistent pattern that is cohesive with the overall theme. Then we should document what pattern we're following so that others contributors know how to follow it when adding or updating rules.

nicorikken:
I did some testing on the `-webkit-font-smoothing`: setting it to 'none' removes it
screen shot 2016-04-16 at 10 43 02
whereas both the 'auto' setting and removing the definition entirely show font smoothing active.
screen shot 2016-04-16 at 10 43 54
So, I agree with removing this line.

ldez:
`-webkit-font-smoothing` is already removed.

nicorikken:
@ldez I thought so, but was misled by the comment by @mojavelinux I'll do a final check and merge your contributions. Nice work!

nicorikken:
@ldez I've merged it. I get the sense you have more experience with styling. Can you provide some reference info to act as a primer into styling in Atom? It could improve this project as well.

ldez:
The primary references:
- http://lesscss.org/functions/

Atom theme variables:
- https://github.com/atom/atom/blob/master/static/variables/syntax-variables.less
- https://github.com/atom/atom/blob/master/static/atom.less
- http://flight-manual.atom.io/hacking-atom/sections/creating-a-theme/#_atom_theme_vars

Some others references:
- https://github.com/atom/styleguide
- https://github.com/atom/template-syntax/blob/master/stylesheets/syntax-variables.less

mojavelinux:
Sorry for causing confusion. :+1: on getting it merged!

nicorikken:
@ldez I've just added your comments in an initial contributing guideline #54  Thanks!

PR #36, opened 2015-10-19
Added single-line comments

blablatros:
This is just a quick fix, I am not that familiar with Atom internals.

* Added a match for single-line comments
* Created `.comment` style that applies to all comments (block and single-line)

### DIFF
--- a/grammars/asciidoc.cson
+++ b/grammars/asciidoc.cson
@@ -85,6 +85,14 @@
     ]
   }
 
+  # Matches single line comments
+  #
+  # // A single-line comment
+  {
+      'match': '^\/\/.*$'
+      'name': 'markup.block.asciidoc'
+  }
+
   # Matches example block
   #
   # Examples
--- a/grammars/asciidoc.cson
+++ b/grammars/asciidoc.cson
@@ -89,7 +89,7 @@
   #
   # // A single-line comment
   {
-      'match': '^\/\/.*$'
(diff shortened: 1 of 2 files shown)


blablatros:
I need to amend this one to match single-line comments that start with whitespace...

nicorikken:
Hi @blablatros Too bad this pull request has been out for so long. I'd like have this feature in my editor, so thanks for your contribution. Let me help speed it up, by giving some feedback. I tested your master branch using the following lines:
```
/ test0
/test1
//
// test2
//test3
/// test4
///test5
```
Which asciidoctor renders to:
`/ test0 /test1 /// test4 ///test5`

Taking this as the truth, it seems that the space trailing the two slashes is not required for detection.
The match would then have to be like: `'match': '^\/{2}.*$'` or your original `'match': '^\/\/.*$'`. Can you explain your change to the trailing whitespace requirement?

Similarly it would be great if you could add some tests (in spec/asciidoc-spec.coffee), so that other contributors can ensure they are not breaking the comment feature. When editing the package in Atom, you can run the current package specs via the command pallet.

If I can help out in any way, please let it know.

mojavelinux:
> Taking this as the truth, it seems that the space trailing the two slashes is not required for detection.

Correct.

nicorikken:
Handled in PR #61 which has just been merged. Therefore I'll close this PR.

Issue #9, opened 2015-04-16
List with bold is not correctly highlighted


alefragnani:
Hi,

When I use a list with bold at the beginning, the bold is not correctly highlighted. Look at the code below:

![print-asciidoc-list-bold](https://cloud.githubusercontent.com/assets/3781424/7171837/40e6c3be-e3b4-11e4-9d42-757e7a8d5a1f.png)

The asciidoc-preview package otherwise, shows correctly.

Thanks

antonmoiseev:
I see, thank you for reporting this! We'll take a look.

nicorikken:
This requires a change in this part of the language definition: https://github.com/asciidoctor/atom-language-asciidoc/blob/master/grammars/asciidoc.cson#L277-L309

nicorikken:
I started working on this issue, learning Ruby regex's on the job. It seems I basically have to exclude the bullet matches, both before and after the bold section. If somebody is eager to help, you can start hacking around at http://rubular.com/ for starters.

nicorikken:
The regex `(?:^|\s|(?:^\s*[\*]{1,5}[\s]))(\*[^\s^\*]+[^\*]*\*)(?:\s|$)` seems to be working fine, although non-capture groups are clearly not the right solution. This will have to be transformed to lookahead and lookbehind statements, although in particular the lookbehind is difficult as not all regex-features are allowed.

mojavelinux:
>  learning Ruby regex's on the job

:+1:  That's how we roll around here ;)

nicorikken:
@mojavelinux Does there happen to be a repository of asciidoc(tor) related regex's?

mojavelinux:
You can find most of them in constants in the following file:

https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor.rb#L398-L1099

nicorikken:
Nailed it! I'll add some of my in-progress-tests and make a pull request. Also got some additional grammar improvements underway as well. `(?<=^|\\s)(\\*[^\\s^\\*][^\\*]*\\*)(?=\\s|$)`

mojavelinux:
Great job!

nicorikken:
I do have one additional test, which does work in the editor, but fails via the spec file. A bit of an edge-case though:

mojavelinux:
I'm not quite sure I understand this test. Is this the sample text?

```
* *Not bold
text*
```

[...]

PR #45, opened 2016-03-24
fix: list with bold highlighting

nicorikken:
In lists with bold content the highlighting now
works correctly. Some edge-cases were added to the
tests to help with future changes.


nicorikken:
Regarding issue 9: https://github.com/asciidoctor/atom-language-asciidoc/issues/9

nicorikken:
When can I expect this to be merged? Because I found some edge-cases involving punctuation marks behind the last asterisk.

mojavelinux:
@antonmoiseev Do you have time to merge? If you need more assistance, we could consider expanding the merge team.

nicorikken:
This pull request can be held for now, as I want to incorporate the formatting definitions as hinted by @mojavelinux in pull request #47.

grammars/asciidoc.cson
```diff
@@ -303,7 +303,7 @@
       #
       #   *bold phrase*
       {
-        'match': '(?<=^|\\s)(\\*[^\\*]+\\*)'
+        'match': '(?<=^|\\s)(\\*[^\\s^\\*][^\\*]*\\*)(?=\\s|$)'
```
mojavelinux: I think the final group should be:

```
(?= |\\t|$)
```

We want to match spaces and tabs but not other forms of whitespace. Technically, the result is probably the same since as far as understand the grammar matches per line, but this communicates the intent better.

nicorikken:
Next fix would be to neglect bullets from the start of the line, as currently bulleted lists are taken as multi-line bold content.

mojavelinux:
Exactly. We need to crack this nut of how to only match blocks at the left margin and other scoping rules. Once we start adding that, this will be the best syntax highlighter for AsciiDoc out there.

nicorikken:
That would involve some regex duplication and nesting. It should just be the logical combination of other cases.

nicorikken:
It seems the highlighting for the last commit works nicely in practice, but the spec still fails, as it tokenizes
`*** Level 3
**`
as bold, even though this is initiated by regular bullets.

mojavelinux:
It seems like scopes may resolve the last example you gave. If we match for a bullet item first, then pass the content to the inline rules, it should stop trying to match on the bullet markers.

mojavelinux:
That's basically how the Asciidoctor parser works. It looks for block-level items, then it parses the content and down and down it goes.

nicorikken:
I've based my next iteration on the nesting. It does seem to fail in the sense that the sidebar block is parsed using the bold inline parsers, and the same goes for the lists. At least, that's what the package specs are saying. The highlighting does however seem to work just fine.

Second to that, I am wondering whether or not I should include the macro (and alike) exclusion which are present at the start of the asciidoctor.rb regexes (like `\\?(?:\[([^\]]+?)\])?`). Maybe proper ordering inside the inline patterns repository could avoid unnecessary complication (and duplication) of regexes in this way.

Please share your thoughts and/or check out this branch to give it a go.

mojavelinux:
> Maybe proper ordering inside the inline patterns repository could avoid unnecessary complication...

I think we definitely need to review the ordering of elements the grammar.

> I am wondering whether or not I should include the macro (and alike) exclusion which are present at the start of the asciidoctor.rb regexes

I'm not sure what you mean by macro exclusion. Do you mean the match of the optional attribute list? We definitely need to start matching the full form, which is:

```
[<attrlist>]`<text>`
```

mojavelinux:
We should definitely looking for blocks before inlines. The grammar file is the reverse of what it should be. If we first match for the blocks, then that should allow the inline matches to be simplified.

nicorikken:
Indeed I meant attributes (still having to get familiar with the advanced features of AsciiDoc). It does relate to the object at hand, but maybe it shouldn't be included in the highlighter, but it should have it's own highlighter. Another option would be to included it in the parsing, and hand it over to a separate highlighter, to provide more context on the attributes.

I'll have to give it some prototyping to see if a reordering will succeed, or if it will fail on other parts. Moving the inline patterns to the end of the main lists should then solve this issue, allowing less complicated regex's.

nicorikken:
I reversed the pattern order, which makes all the tests succeed :smile:
Now I can reintroduce the attribute neglecting regex-parts, and clean it all up.

mojavelinux:
> Indeed I meant attributes (still having to get familiar with the advanced features of AsciiDoc).

Don't worry, I'm still trying to clarify all the terminology. We're both learning.

> Another option would be to included it in the parsing, and hand it over to a separate highlighter, to provide more context on the attributes.

I think this is the right choice. We should match the whole expression, then pass the attribute list part to a subpattern to parse out the list. At the moment, the only syntax permitted for inlines are `[role]`, `[role1 role2]` or `[.role#id]`. In other words, only roles and ids atm, not arbitrary key/value pairs.

> Now I can reintroduce the attribute neglecting regex-parts, and clean it all up.

\o/

nicorikken:
Just added the `asciidoctor.rb` based attribute regex. Will have to handle the capture groups the right way. @mojavelinux then I'll capture it and pass it on.

mojavelinux:
:+1:

Now that we know we are using the [oniguruma regex engine](https://github.com/atom/node-oniguruma), I wonder if we can simplify the following expression:

```
[\\S\\s]
```

to

```
.
```

The `[\S\s]` is a trick in the JavaScript regex engine to match all characters across all lines. We also know that `match` only looks at one line at a time, making it even more unnecessary.

nicorikken:
Selection works, now I'll have to dispatch it to the associated pattern. Documentation in the Asciidoctor manual: http://asciidoctor.org/docs/user-manual/#setting-attributes-on-an-element There seem to be multiple regexes related to style attributes (quotes, admonition), but none specific to roles. Am I overlooking something?

mojavelinux:
The documentation about attributes somewhat misleading at the moment. Attribute lists really only apply to blocks, block macros and inline macros. What the attribute list on formatted text accepts is very different at the moment. It only accepts a list of roles and/or an id. The documentation needs to be updated to reflect this.

For now, I wouldn't worry about parsing the attribute list on formatted text. I just recommend wrapping a scope around the attribute list and call it a day for now.

spec/asciidoc-spec.coffee
```diff
@@ -16,11 +16,142 @@ describe "AsciiDoc grammar", ->
...
+
+    it "when text is '*bold text*'", ->
+      {tokens} = grammar.tokenizeLine("*bold text*")
+      expect(tokens.length).toEqual 1
+      expect(tokens[0]).toEqual value: "*bold text*", scopes: ["source.asciidoc", "markup.bold.asciidoc"]
+
+    it "when text is '*bold*text*'", ->
+      {tokens} = grammar.tokenizeLine("*bold*text*")
+      expect(tokens.length).toEqual 1
+      expect(tokens[0]).toEqual value: "*bold*text*", scopes: ["source.asciidoc"]
```
nicorikken: It seems this should render *bold*text* als bold, but somehow it doesn't. I'll have to look into this.

nicorikken:
@ldez I've included your bold-tests from your PR #51

nicorikken:
If others agree, I'd like to merge this for now, aiming to fix the `*bold*text*` and multi line tests at a later stage.

grammars/asciidoc.cson
```diff
@@ -291,21 +285,41 @@ patterns: [
...
+      }
+
       # Matches bold constrained phrases
       #
       # Examples
       #
       #   *bold phrase*
       {
-        'match': '(?<=^|\\s)(\\*[^\\*]+\\*)'
+        # asciidoctor.rb: /(^|[^\w;:}])(?:\[([^\]]+?)\])?\*(\S|\S[\s\S]*?\S)\*(?![\w])/m
```
ldez: You have a forgotten code comment.

grammars/asciidoc.cson
```diff
@@ -291,21 +285,41 @@ patterns: [
...
   'inline':
     'patterns': [
 
+      # Matches bold unconstrained phrases
+      #
+      # Examples
+      #
+      #   b**old** phrase
+      {
+        # asciidoctor.rb: /\\?(?:\[([^\]]+?)\])?\*\*([\s\S]+?)\*\*/m
```
ldez: Same thing : forgotten comment

ldez:
There is many TODO, you are sure to want merge with those comments ?

nicorikken:
@ldez I kept the asciidoctor.rb comments intentionally, but for cleanness I can of course remove them. Regarding the remaining TODO's, the multi-line requires a more extensive rewrite (begin and end). The ´*bold*text*´ is something that should be working for this match version as well. Not sure about merging. I can of course keep working on this issue in this branch, getting it to a better state. It's more that I've satisfied the initial goals.

ldez:
Add comments and todo on main branch is not a good idea I think.
Merge a clean PR (without comments and todo) as a baby step seems a good idea.

mojavelinux:
I think it's safe to remove the comments about asciidoctor.rb. What's most useful are the examples.

I noticed from the javascript language grammar file that there is a space for a comment in the definition. Perhaps we could use that to describe what is being matched. (See https://github.com/atom/language-javascript/blob/v0.118.0/grammars/javascript.cson#L30)

nicorikken:
@ldez Alright, I'll bite the bullet and perfect this branch before merging then.
@mojavelinux Using the comment feature seems like a good practice, making it machine parsable, say for generating test-cases.
I just added and perfected the unconstrained fix as suggested in #64 Now I need to doe the same for the constrained match.

nicorikken:
This final commit should do the job. Test-cases added, cleaned up, and TODO list elements added, as I ran into a regex issue with the associated asterisk. How is this for a PR?

grammars/asciidoc.cson
```diff
@@ -7,17 +7,11 @@ fileTypes: [
...
 ]
 
-patterns: [
-
-  # Includes inline patterns, see repository section below.
-  {
-    include: '#inline'
-  }
-
+'patterns': [
```
ldez: you can remove single quote here

grammars/asciidoc.cson
```diff
@@ -291,21 +300,47 @@ patterns: [
     'captures':
       '1': 'name': 'markup.list.bullet.asciidoc'
   }
+
+  # Includes inline patterns, see repository section below.
+  {
+    'include': '#inline'
```
ldez: if you want to move this include, put this include in the repository

nicorikken: What do you mean here? Because if I put the `include` in the `repository`, it will be a cyclical reference.

ldez: Sorry forgot my remark. I've bugging. My minds work in "generator mode" :wink:

ldez: But why you have move this include ?

nicorikken: Good for you. I have to get used to 'generator mode'. Glad it isn't as nested and badly named as Ansible deploy definitions.
I've moved the include as reordering was the most neat and convenient way to get around the issue of bulleted lists getting recognized as bold content. Negatively matching the bullets of the list results in a very nasty regex, which is hard to debug. The required complexity was over my head at least. So by reordering, the lists are matched first, and second bold content is located. This works beautifully, and I haven't discovered any breakage as a result of the reordering.

ldez:
I think you split tests files in multiple files like that:
- `spec/asciidoc-bold-spec.coffee`
- `spec/asciidoc-spec.coffee`

nicorikken:
PR replaced by PR #69

nicorikken:
PR #69 was merged, so this one can be closed.

mojavelinux:
Absolutely incredible @ldez and @nicorikken! The AsciiDoc language support in Atom just :rocket: off!

Issue #21, opened 2015-07-29
Italic is not properly highlighted


antonmoiseev:
Underscore inside a word breaks highlighting:
![chapter3_adoc_-__users_amoiseev_code_git_manning_com_agileauthor_fain_drafts_-_atom](https://cloud.githubusercontent.com/assets/182853/8961526/61702cf8-361e-11e5-8086-3ab1da5c47e8.jpg)

nicorikken:
I'm pretty sure I can reuse my fix for the bold highlighting in issue #9. I'll see what I can do.

nicorikken:
Fix implemented, pull-request created.

nicorikken:
This particular issue was fixed in release 1.0.0, even though the italic highlighting in itself can be improved. It can now be too greedy. I'll however close this specific issue.

PR #46, opened 2016-03-24
fix: special chars in block titles

nicorikken:
This issue will fix special characters breaking block title highlighting.

### DIFF
--- a/spec/asciidoc-spec.coffee
+++ b/spec/asciidoc-spec.coffee
@@ -173,13 +173,12 @@ describe "AsciiDoc grammar", ->
 
   it "tokenizes block titles", ->
     {tokens} = grammar.tokenizeLine("""
-                                    .An example example
+                                    .An e-xample' e_xample
                                     =========
                                     Example
                                     =========
                                     """)
-    expect(tokens[1]).toEqual value: "An example example", scopes: ["source.asciidoc", "markup.heading.blocktitle.asciidoc"]
-    expect(tokens[3]).toEqual value: "=========", scopes: ["source.asciidoc", "markup.block.example.asciidoc"]
+    expect(tokens[1]).toEqual value: "An e-xample' e_xample", scopes: ["source.asciidoc", "markup.heading.blocktitle.asciidoc"]
 
   it "tokenizes Mardown-style headings", ->
     {tokens} = grammar.tokenizeLine("# Heading 0")


grammars/asciidoc.cson
```diff
@@ -48,8 +48,7 @@
   # ========================
   #
   {
-    'begin': '^\\.(\\w[\\w ]*\\w)'
-    'end': '$'
+    'match': '(?<=^\\.)(\\w.+)(?=$)'
```
mojavelinux: There is another PR also proposing to fix this issue. See #48.

In that PR, I describe the correct way to match a block title.

```
'begin': '(?<=^\\.)([^\\s.].*)'
```

In other words, it should match a title that is at least one character in length and does not start with a space or dot.

nicorikken:
Based on the asciidoctor.rb source, and taking in my learnings in the italic highligh ( #21 ), I've now come up with two matches, for both constrained and unconstrained phrases:

Constrained: `'match': '(^|[^\\w;:}])(?:\\[([^\\]]+?)\\])?(?!\\*{3})(\\*\\S\\*|\\*\\S[\\s\\S]*?\\S\\*)(?![\\w])'`
Unconstrained: `'match': '\\\\?(?:\\[([^\\]]+?)\\])?(\\*\\*[\\s\\S]+?\\*\\*)'`

It needs some more testing and unit tests, but for now this seems to be the right solution.

mojavelinux:
Excellent! However, I think your comment applies to #21, not this one.

@nicorikken You've been doing excellent work on this package and we'd like to invite you to become a "committer". That will give you the authority to merge your pull requests once we decide they're ready to go, saving you the trouble of having to wait on anyone else to merge.

mojavelinux:
I think we should merge #48 when it is ready, then rebase this one and proceed to include the test and perhaps a refinement to the regex.

nicorikken:
@mojavelinux Thank you very much for the invite. I'll happily help to handle the PRs. I'll probably need some guidance and help in handling especially the first few PRs, to get familiar with Asciidoctor practices and the ins and outs of git merging in general. But I'll get back to you on that then.

mojavelinux:
Excellent! Of course, we are here to help you! The most important thing is that we start to integrate your great work.

nicorikken:
#48 is merged, and now indeed tests can be included. For now I have no reason to doubt the included regexes.

mojavelinux:
I think this PR needs to be rebased. Then we can assess where we stand.

nicorikken:
@mojavelinux do you agree on merging this changed test-case?

mojavelinux:
Yes indeed!

Technically, we could consolidate the rule so that it only uses the begin key:

```
  {
    'begin': '^\\.([^\\s.].*)$'
    'captures':
      '1': 'name': 'markup.heading.blocktitle.asciidoc'
  }
```

But that's effectively the same as what it is now. It's up to you if you'd like to consolidate it.

nicorikken:
The consolidation scrambles up my parsing. I guess a match would do likewise. I'll leave it as merged for now.

nicorikken:
Merged!

mojavelinux:
Thanks for checking!

:+1:

PR #47, opened 2016-03-24
fix: italic highlight with underscores

nicorikken:
Using a lazy quantifier and a lookahead to ensure
that the highlight will break at the first
underscore followed by a space or line end.

Fixes issue #21.

### DIFF
--- a/grammars/asciidoc.cson
+++ b/grammars/asciidoc.cson
@@ -297,7 +297,7 @@
   'inline':
     'patterns': [
 
-      # Matches bold pharses
+      # Matches bold constrained phrases
       #
       # Examples
       #
@@ -308,13 +308,24 @@
           '0': 'name': 'markup.bold.asciidoc'
       }
 
-      # Matches italic pharses
+      # Matches italic constrained phrases
       #
       # Examples
       #
       #   _italic phrase_
       {
-        'match': '(?<=^|\\s)(_[^_^\\s].*?_)(?=\\s|$)'
+        'match': '(?!_{4,}\\s*$)(?<=^|[^\\w;:])(?:\\[(?:[^\\]]+?)\\])?(_\\S_|_\\S[\\s\\S]*?\\S_)(?!\\w)'
+        'captures':
+          '0': 'name': 'markup.italic.asciidoc'
+      }
+
+      # Matches italic unconstrained phrases
+      #
+      # Examples
+      #
+      #   i__talic__ phrase
+      {
+        'match': '\\\\?(?:\\[([^\\]]+?)\\])?(__(?!_)[\\s\\S]+?__)' # now including the underscores in the match
         'captures':
           '0': 'name': 'markup.italic.asciidoc'
       }
--- a/grammars/asciidoc.cson
+++ b/grammars/asciidoc.cson
@@ -314,7 +314,7 @@
       #
       #   _italic phrase_
       {
-        'match': '(?<=^|\\s)(_[^_]+_)'
+        'match': '(?<=^|\\s)(_[^_^\\s].*?_)(?=\\s|$)'
         'captures':
           '0': 'name': 'markup.italic.asciidoc'
       }
--- a/spec/asciidoc-spec.coffee
+++ b/spec/asciidoc-spec.coffee
@@ -28,6 +28,12 @@ describe "AsciiDoc grammar", ->
     expect(tokens[1]).toEqual value: "_italic_", scopes: ["source.asciidoc", "markup.italic.asciidoc"]
     expect(tokens[2]).toEqual value: " text", scopes: ["source.asciidoc"]
 
+  it "tokenizes _italic_ text with underscores", ->
+    {tokens} = grammar.tokenizeLine("this is _italic_text_ with underscores")
+    expect(tokens[0]).toEqual value: "this is ", scopes: ["source.asciidoc"]
+    expect(tokens[1]).toEqual value: "_italic_text_", scopes: ["source.asciidoc", "markup.italic.asciidoc"]
+    expect(tokens[2]).toEqual value: " with underscores", scopes: ["source.asciidoc"]
+
   it "tokenizes HTML elements", ->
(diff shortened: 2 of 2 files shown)


mojavelinux:
>  the highlight will break at the first underscore followed by a space or line end.

That's not the correct rules for constrained (single char) formatting marks. Rather, they match when the inside of the mark is a word character and the outside of the mark is a non-word character. This is now described in the user manual. See http://asciidoctor.org/docs/user-manual/#formatting-marks

nicorikken:
@mojavelinux, thanks for the pointer, I learned something new about the Asciidoc syntax. I'll improve this fix. Stay tuned...

nicorikken:
Having read the manual, my current WIP is the following match string:

nicorikken:
Based on the formatting remarks I've come up with this new matching statement:
`'match': '(?<=^|\\W)(?<!;)(_(?!\\s|_).*?(?<!\\s|_)_)(?=\\W|$)'`
It also covers the use-case of a semi-column in front of the first underscore, as stated as a case in the formatting remarks.

I'm still struggling with multi-line matches in my editor, although the tests seem to run just fine. Maybe some packages conflict.

Also I started working on the unconstrained italic quotes, which has the potential of conflicting with quoting blocks and the constrained italic quotes.

nicorikken:
And just for fun, anyone can try this out: http://rubular.com/r/ohuNwqN5z2

nicorikken:
The main asciidoctor.rb regexes are good for reuse, although they do not include the underscores in the capturing (which makes sense from a rendering point of view). Tinkering with it, I've now come up with the following expressions:

Constrained phrases:
`'match': '(?!_{4,}\\s*$)(?!^|[^\\w;:])(?:\\[(?:[^\\]]+?)\\])?(_\\S_|_\\S[\\s\\S]*?\\S_)(?!\\w)'`

Unconstrained phrases:
`'match': '\\\\?(?:\\[([^\\]]+?)\\])?(__(?!_)[\\s\\S]+?__)'`

Through daily use it seems I've ironed out all glitches. I'll include these cases in the tests in the to-be update.

nicorikken:
I made a small error in the constrained phrases. Current WIP regex:
`'match': '(?!_{4,}\\s*$)(?<=^|[^\\w;:])(?:\\[(?:[^\\]]+?)\\])?(_\\S_|_\\S[\\s\\S]*?\\S_)(?!\\w)'`

mojavelinux:
Amazing work @nicorikken! That's at least as good if not better than the vim syntax, which is a good first milestone to best. For reference: https://github.com/dagwieers/asciidoc-vim/blob/master/syntax/asciidoc.vim and https://github.com/dagwieers/asciidoc-vim/blob/master/asciidoc-testcases.txt.

@antonmoiseev do you need help reviewing and merging pull requests?

nicorikken:
Ah nice, even more resources to use during development. With the latest commit 536e85a added to this branch, now the italic should work just fine. I'm glad to receive any comments regarding remaining work to get this code merged.

mojavelinux:
As I mentioned in the other issue, we're inviting you to become a committer on this repository, which will allow you to merge PRs once we decide they are ready. You first need to accept the invitation at https://github.com/asciidoctor.

mojavelinux:
I tested this and it works great. I'm giving it my :+1: for merge.

nicorikken:
After some final tests, I've merged my commit :tada: Now I can take this experience regarding regex into other issues like bold highlights.

ldez:
Sorry but your changes break CoffeeLint rules and this failed my PR #52.
I'm going to fix the breaking rules, in my PR.
Please merge the #52 to avoid that.

nicorikken:
(Facepalm!) I guess that's why #52 is needed. I'll get some fixes in.

nicorikken:
@ldez Apart from merging #52 I intent-fully included the spaces, to include a particular test-case. Or am I wrong in assuming that the spaces influence the CoffeeScript tests? If they do not influence the tests, then I can leave them out. Otherwise I like to leave them in, but that would conflict with having CoffeeLint rules handled by Travis. Can you provide some context on CoffeeLint and this issue?

ldez:
`no_trailing_whitespace` is a good default rule.
If you want to preserve spaces you must replace block strings by a simple line with `\n`.

I fix my PR by changing tests cases for preserve spaces.
Are you ok ?

ldez:
or I can use exclusion in specific line.

```
# coffeelint: disable=no_trailing_whitespace
{tokens} = grammar.tokenizeLine("""
                                this is _multi-
                                line italic_ text
                                """)
                                # coffeelint: enable=no_trailing_whitespace
```

What do you prefer ?

ldez:
I think inlining it's a better way:
- lot of editor (like Atom) use [Beautifier tool](https://atom.io/packages/atom-beautify), it will automatically remove trailling-space. (I use this tool)
- manually exclude induce contributors can abuse of this option

nicorikken:
@ldez Then single line strings with a `\n` would be the way to go. I use the [trailing-spaces](https://github.com/wpillar/atom-trailing-spaces) package so achieve a similar result. So my trailing spaces certainly were intended. In retrospect I'm not sure if the spaces bring anything to the test-cases. Only very strange edge-cases I guess. So, getting back to your original fix (stripping whitespace) seems to be the best solution for now. Either way thanks for all the info and explanations.

ldez:
@nicorikken it's done! :smile:

mojavelinux:
Great job @nicorikken and @ldez! I look forward to the improvements this change is going to bring.

I have one follow-up comment. I think the unconstrained match should be listed first. The reason is that is the order in which the Asciidoctor parser processes the matches. Assuming the grammar is matched in document order, it should be listed first. Can you switch them around?

nicorikken:
This pull request still has some failures that need to be resolved. Just consider this input string:
`_italic_ I __italic_ I ___italic_ I ___italic__ I ___italic___ I __italic___ I _italic___`
Some of the I's are italic, and so can the related underscores be. The regex was too greedy.

PR #48, opened 2016-03-25
Refine grammar definition for generic inline/block macro and block title

ProgramFan:
Current language-asciidoc has the following flaws:
1. generic inline macros like `cite:[np]` are not properly reconganized.
2. generic block macros like `bibliography::[]` are not properly reconganized.
3. block title with '-' such as '.Real-world applications' are reconganized as '.Real'.

This pull request fixes the above problems by refine the grammar definition regex. It directly enables syntax highlighting support for asciidoctor-bibtex introduced macros, namely `cite:[key]` and `bibliography::[]`, and fixes block titles with '-' in words.


grammars/asciidoc.cson
```diff
@@ -48,7 +48,7 @@
   # ========================
   #
   {
-    'begin': '^\\.(\\w[\\w ]*\\w)'
+    'begin': '^\\.(\\w.*\\w)'
```
mojavelinux: To be accurate, the first character should be matched as:

```
[^\s.]
```

Which means "not space or period".

There is no need to match a trailing word character. Thus, the whole thing can be shorted to:

```
'begin': '^\\.([^\\s.].*)'
```

mojavelinux:
My understanding is that this change allows short macros (macros with a target) to be matched. Is that a correct assumption? If so, then this change is definitely something we want. It should also allow `toc::[]` and `kbd:[Ctrl,T]` to be matched then.

grammars/asciidoc.cson
ProgramFan: I agree. I am just wonder why the original matches use `\w`. Are there special considerations?

mojavelinux: The plugin began as an MVP and, in the essence of time, the original line
was overly simplified. Now we just need to make it accurate.

ProgramFan:
@mojavelinux Exactly.

ProgramFan:
I have mistakenly created the pull request to merge the master branch, where I also committed some improvements in the snippets, where I changed the header styles and the block deliminaters according to recommanded asciidoc style. If this is not valid for a merge, I can revert these changes.

mojavelinux:
I would prefer if you did these as two separate pull requests. Not only does it allow us to review the changes individually, you also get credit for both changes in the history (two entries).

ProgramFan:
I have made a seperate PR for snippts and clear the CI errors. Would you please review it for possible merge? @mojavelinux

grammars/asciidoc.cson
```diff
@@ -48,7 +48,7 @@
   # ========================
   #
   {
-    'begin': '^\\.(\\w[\\w ]*\\w)'
+    'begin': '^\\.([^\\.\\s].*)'
```
mojavelinux: This should be:

```
'begin': '^\\.[^\\s.].*)'
```

It's not necessary to escape a dot in a character group.

mojavelinux:
I have a minor change request. Once you address that, we can move forward.

ProgramFan:
Done.

mojavelinux:
Thank you!

@nicorikken I think this one is ready to go. Can you review? When you merge, please use "squash and merge" so we get this down to one commit.

nicorikken:
I have merged the current state of this pull-request. I've run some basic tests on the examples given in the documentation and this tread. @ProgramFan can you provide some test-cases which previously failed and with this commit work just fine? In this way the set of tests can be expanded to cover the relevant edge-cases. Either way, thanks for this improvement.

ProgramFan:
I want to, but how can I add it? in the spec?

nicorikken:
You can add specs in the file https://github.com/asciidoctor/atom-language-asciidoc/blob/master/spec/asciidoc-spec.coffee They describe an input line and test the expected outcome styles. You can look at the other specs for inspiration. You can run the specs from the command pallet.

nicorikken:
@ProgramFan I've rebased my PR #46 leaving an improved spec on the block titles. I'd still like some macro tests.

PR #49, opened 2016-03-31
Improve snippets

ProgramFan:
Update snippets to match recommended style from asciidoctor.org


mojavelinux:
:+1: :100:

ProgramFan:
Is it ready for a merge? If not, I am will to make corrections. :smile:

nicorikken:
This section of the plugin is new for me, interesting though. Thanks for improving my editor as well! :+1:

nicorikken:
@ProgramFan I just checked out out your branch, and it works great. Up until now I had never used the snippets, but especially for blocks it can be powerful. I agree with the header commit, because the trailing marks are nasty and unnecessary. Remove the long block separates is great likewise. I'm however not as sure about the Author block. Normally I use the author block in a different way. Can you elaborate on the style? Maybe I'm just doing it wrong :stuck_out_tongue: 
```
= Test document
:Date: 25-03-2016
:Author: Nico Rikken <nico@domain.com>
```
As a sidenote: I just found your other contributions and repo's related to Asciidoc. Great! :+1: Lots of interesting tools to play with in the future. :thought_balloon: { If only I would have used Asciidoc for my graduation paper and report instead of LaTeX }

ProgramFan:
@nicorikken The author block is defined as the optional author and revision line of document title, as defined by asciidoc in http://asciidoctor.org/docs/asciidoc-writers-guide/#the-document-header. It's widely used in asciidoctor.org documents. Your author block style is also valid (it's shall be called 'author' variables instead of author block, though.), but the asciidoctor one is more widely used so better for the snippet.

I am currently using asciidoc for the preperation of technical papers (academic paper, research reports etc, in Computer Science). I find asciidoc much easier than latex. I usually prepare content-ready drafts using asciidoc and convert to latex using asciidoctor-latex or convert to word documents using asciidoctor and pandoc. This indeed saves a lot of time since I do not need to pay attention to formatting issues in the preparation stage. This is why I created asciidoctor-bibtex and contributed to asciidoctor-mathematical. I hope they are useful to you, but just remember that **Content IS THE KING**.

nicorikken:
@ProgramFan Referring to the asciidoctor writers-guide certainly makes sense. Even the Asciidoc user guide explains that the document header like you described, equates to setting the equivalent document attributes. Therefore I agree, focusing on the document header makes sense. So with that out of the way, I opt for this pull request to be accepted as well.

Content certainly is king, that's why I love Asciidoc(tor) so much, it allows me to focus on the content, much more than on the rendering details. In that regard LaTeX isn't much of a markup language. Having BibTeX integration is a nice feature, especially since it allows integration with all the BibTeX tooling.

mojavelinux:
The preferred header style is:

```
= Test document
Nico Rikken <nico@domain.com>
2016-03-25
```

You can use explicit attribute names (author, email and revdate), but the shorthand is what is generally used.

mojavelinux:
I'm ready to merge this. @antonmoiseev do approve the merge?

nicorikken:
@mojavelinux Am I allowed to merge this pull request? I have reviewed and tested it.

mojavelinux:
Yes please.

Now that you are a co-maintainer, you can choose to merge when you think it is time. I generally like to keep pull requests open for review for 1-2 days and get at least one person to do a code review. These are not strict rules, but generally lead to a better outcome.

mojavelinux:
I'm thinking in the future we might want to drop the shorthand down to a single letter plus the delimiter characters. Something like `b==` (which is currently `bl==`). Keep that in mind when you are using this feature and pay attention to what would feel more comfortable.

nicorikken:
@ProgramFan Thanks for the contribution, it has just been merged. You should see it pop up in your editor when we bump up the version number. Although I assume it is part of your editor setup already ;)

ProgramFan:
Thanks for you effort, @nicorikken @mojavelinux. I am glad to contribute, since asciidoctor is such a fatastic tool.

mojavelinux:
Thanks to you both, @ProgramFan and @nicorikken!

PR #51, opened 2016-04-16
fix: bold and italic

### LINKED ISSUE #41: bold in bullet does not highlight correctly
In a bullet like this:

    * Here is some text *now I want bold*.

the highlight begins at the first * and ends at the second *.  The preview, however, is correct.

The highlighter should understand that a * followed by a space at the start of a line is a bullet, unless there is exactly one other * anywhere else in the line. (Assuming no escaped or literal *).

ldez:
Before fix:

before

After fix:

after

Fixes #41 and #9

### DIFF
--- a/grammars/asciidoc.cson
+++ b/grammars/asciidoc.cson
@@ -297,24 +297,24 @@
   'inline':
     'patterns': [
 
-      # Matches bold pharses
+      # Matches bold phrases
       #
       # Examples
       #
       #   *bold phrase*
       {
-        'match': '(?<=^|\\s)(\\*[^\\*]+\\*)'
+        'match': '(?<=^|\\s)(\\*[^ \\*].+?\\*)(?= |\\t|$)'
         'captures':
           '0': 'name': 'markup.bold.asciidoc'
       }
 
-      # Matches italic pharses
+      # Matches italic phrases
       #
       # Examples
       #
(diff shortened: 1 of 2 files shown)


ldez:
Sorry I don't see #9 before create the PR.

mojavelinux:
Are there any tests you don't think are covered by those other pull requests? I would hate to lose good tests. Perhaps we can wait until those are merged, then see what isn't covered at that point.

mojavelinux:
#45 seems to miss this case:

```
*bold *text*
```

mojavelinux:
#47 seems to cover all the italic cases fine. I think #45 just needs to be updated to match the fixes in #47.

nicorikken:
@ldez I haven't yet given this a proper look, but I believe I should. My italic PR #47 turned out to be too greedy, which needs a fix. Furthermore my bold fix #bold #45 is still having some issues in the spec-area. Even though the rendering itselfs seems to be in order :confused:

nicorikken:
@ldez I gave your bold regex a shot, but it fails on special characters outside the asterisks. Please check my progress on #45 for more info.

ldez:
I close this PR.

nicorikken:
Nice testcases though, and added to that I like the grouping using the `describe`. Something to include for the other tests as well.

PR #55, opened 2016-04-17
feat: extends supported languages for code block

ldez:
- dockerfile
- properties (git-config, java-properties)
- makefile
- perl and perl6
- toml

### DIFF
--- a/grammars/asciidoc.cson
+++ b/grammars/asciidoc.cson
@@ -530,7 +530,7 @@
             'end': '\\1'
             'endCaptures':
               '0': 'name': 'support.asciidoc'
-            'contentName': 'source.go'
+            'contentName': 'source.embedded.go'
             'patterns': ['include': 'source.go']
           }
         ]
@@ -583,7 +583,7 @@
             'end': '\\1'
             'endCaptures':
               '0': 'name': 'support.asciidoc'
-            'contentName': 'source.gfm'
+            'contentName': 'source.embedded.gfm'
             'patterns': ['include': 'source.gfm']
           }
         ]
@@ -610,7 +610,7 @@
             'end': '\\1'
             'endCaptures':
               '0': 'name': 'support.asciidoc'
-            'contentName': 'source.yaml'
+            'contentName': 'source.embedded.yaml'
             'patterns': ['include': 'source.yaml']
           }
         ]
@@ -637,7 +637,7 @@
             'end': '\\1'
             'endCaptures':
               '0': 'name': 'support.asciidoc'
-            'contentName': 'source.coffee'
+            'contentName': 'source.embedded.coffee'
             'patterns': ['include': 'source.coffee']
           }
         ]
@@ -664,7 +664,7 @@
             'end': '\\1'
             'endCaptures':
               '0': 'name': 'support.asciidoc'
-            'contentName': 'source.json'
+            'contentName': 'source.embedded.json'
             'patterns': ['include': 'source.json']
           }
         ]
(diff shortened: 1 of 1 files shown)


grammars/asciidoc.cson
```diff
@@ -778,6 +778,169 @@
...
+            'patterns': ['include': 'source.perl6']
+          }
+        ]
+      }
+
+      # Matches Toml AsciiDoc code blocks
+      #
+      # Examples
+      #
+      #   [source,rust]
```
nicorikken: Typo?

ldez: fixed!

nicorikken:
All the source definitions seem to exist based on some GitHub code searches. Otherwise I find it hard to check. Maybe the `begin': '(^\\s*-{4,}\\s*$)'` regex can be put in a repository for reference, based on DRY. Either way the code-blocks involve a lot of unavoidable duplication.
Having the typo fix, this seems mergeable. Nice work! :+1:

ldez:
I give you my trick for the languages support.

To get a list of all scope names registered in your Atom instance:

- open the Developer Tools Console: `Ctrl+Shift+I`
- execute the following:

```javascript
Object.keys(atom.grammars.grammarsByScopeName).sort().join('\n')
```

ldez:
And the GitHub GFM languages code (base on [Ace](https://ace.c9.io/)): https://github.com/github/linguist/blob/master/lib/linguist/languages.yml

nicorikken:
Golden!!!! Only one solution: `> CONTRIBUTING.adoc`

mojavelinux:
> the code-blocks involve a lot of unavoidable duplication.

Is it possible use a match group in contentName and patterns fields? If so, we could consolidate these definitions down to a single grammar rule.

If that's not possible, perhaps we can create a script that generates this part of the file. The script would still need to be executed manually, but at least it would allow us to maintain a single rule and have all the permutations generated automatically.

mojavelinux:
I think we should align with the Markdown grammars by using the content name `source.embedded.%s` where `%s` is the language name. wdyt? (perhaps a separate PR)

mojavelinux:
> Golden!!!! Only one solution: > CONTRIBUTING.adoc

I like your thinking!

ldez:
The current `contentName` with `source.%s` create a misleading with specific tools (like [linter-shellcheck](https://atom.io/packages/linter-shellcheck)).
Change to `source.embedded.%s` will fix this issue.

mojavelinux:
Let's do it then!

ldez:
It's already do! :wink:

ldez:
once this PR merge, I'll do a big review on 'code block' grammar.

mojavelinux:
Oops, I missed that. Super!

Let's proceed. :+1:

nicorikken:
This new way indeed confirms to the [GitHub flavored markdown way](https://github.com/atom/language-gfm/blob/master/grammars/gfm.cson). Does this disable the language-specific linters in the code-blocks? This would certainly be helpful, because a typo in the JSON code-block has messed up my whole grammar parsing in the past. I'll get this merged next.

ldez:
Linter are always working but some linters parse whole document, not only the code block, with the current naming.
I've prepare a PR with all languages blocks rewriting, I'm waiting  the merge of this PR.

nicorikken:
I did a small check by checking for the embedded property in the cursor scope. @ldez thanks for the improvement

mojavelinux:
@nicorikken & @ldez, welcome to the [contributors page](http://asciidoctor.org/contributors/) on asciidoctor.org! I'm so glad to have you as part of the team!

PR #57, opened 2016-04-18
refactor: languages support

ldez:
- review regex
- add `name` attribute
- asciidoctor syntax and mardown syntax
- add SASS support
- fix several language bug
- **add AsciiDoctor comment support**

### DIFF
--- a/.coffeelintignore
+++ b/.coffeelintignore
@@ -1,0 +1,1 @@
+spec/fixtures
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+node_modules
--- a/grammars/asciidoc.cson
+++ b/grammars/asciidoc.cson
@@ -462,7 +462,7 @@
       #   }).listen(1337, '127.0.0.1');
       #   ----
       {
-        'begin': '^\\[source,\\s*(javascript|js)\\]$'
+        'begin': '^\\s*\\[source,\\s*(?i:(javascript|js))\\]$'
         'beginCaptures':
           '0': 'name': 'support.asciidoc'
         'end': '^\\s*$'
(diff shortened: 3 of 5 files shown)


nicorikken:
You know PRs are serious when GitHub's UI is giving up :+1:  I'll have a look.

nicorikken:
@ldez Can I conclude the previous regexes were case-sensitive and you fixed that? And by referencing the block start in the end, you force the regex to work on balanced block separators only?

ldez:
Yes your conclusion are good.
I also permit to add space before code block.

ldez:
The main feature is the `name` attribute: allow to add asciidoctor comment scope.

nicorikken:
So this implies that adding `// comment` in a code block will not influence the highlighting within the code block? (preparing for future feature in this way)

ldez:
Not really: now you can press `ctrl+shift+/` in an AsciiDoctor document for comment a line.
And with the attribute `name` I can exclude that pressing those keys add a AsciiDoctor comment style in code block.
Comment in code block respect specific language comment style.

nicorikken:
Doing a simple check on the effect of commenting inside code blocks, it seems that the commenting conforms to the language present in the code block. Testing inside an `[source, sh]` block triggers a `#` comment rather than a `//` comment. Am I missing the point?

ldez:
It's a misunderstood of my part: the descriptor, in `settings/asciidoc.cson` with `not` part, affect `softWrap` option but not the comment scope.

ldez:
I've made a language sample document, do you want I had this file to the project?

nicorikken:
Sample document as in a .adoc file, or in specs. Either way including them will help. When including in tests, we need to watch out for the reduced language support in say travis, contrary to our massive Atom setups which presumably support some exotic languages.

ldez:
I've added the file.
You can see some languages support bug, but I think it's a related to others languages plugins bugs.

nicorikken:
Oh my, nice one! I count and can confirm 3 bugs, in Properties, PHP, and CSS. It seems that not all styles were converted to include `.embedded.` Is this due to rendering issues? To me this seems mergeable.

mojavelinux:
:+1:

nicorikken:
Thanks @ldez  I have merged this wonderful commit. @mojavelinux I've kept both separate commits, as they contribute distinct features (the refactor and the test/doc).

mojavelinux:
:+1:

PR #58, opened 2016-04-21
refactor: code block scope

ldez:
- [x] all the code blocks can be immediately followed by other grammar.
- [x] fix bug with CSS  code block
- [x] add TypeScript language support
- [x] split SCSS to SASS and SCSS
- [x] add custom 'properties' grammar

**Note:**
I use `grammar.tokenizeLines` in the tests because `grammar.tokenizeLine` doesn't work properly with multi-lines content.
- `grammar.tokenizeLine` give false positive with multi-lines because it parse multi-lines as a single line.

Fix #20, #39,  #59

PR #60, opened 2016-04-24
Porting AsciiDoctor regex for monospace

### LINKED ISSUE #38: Inline code blocks not highlighted inside parentheses
example:

```asciidoc
// this one gets highlighted
`inline code`

// this one does not
(`inline code`)
```
I'd say the regex [here](https://github.com/asciidoctor/atom-language-asciidoc/blob/3abc02ee7848c1229846cbeabd6583a7bb928abd/grammars/asciidoc.cson#L373) should probably be converted from:
```
(?<=^|\\s)`[^`]+`
```
to:
```
(?<=^|\\W)`[^`]+`
```

Please say if you'd like me to open a pull request with the changes

ldez:
Porting the https://github.com/asciidoctor/asciidoctor/blob/v1.5.4/lib/asciidoctor.rb#L1177-L1178 to the grammar.

fix #38

### DIFF
--- a/grammars/asciidoc.cson
+++ b/grammars/asciidoc.cson
@@ -381,8 +381,9 @@
       #
       #   `Text in {backticks}` renders exactly as entered, in `monospace`.
       {
-        'match': '(?<=^|\\s)`[^`]+`'
-        'name': 'markup.raw.inline.asciidoc'
+        'match': '(^|[^\\p{Word};:"\'`\\}])((?:\\[([^\\\\]]+?)\\])?`(\\S|\\S.*?\\S)`(?![\\p{Word}"\'`]))'
+        'captures':
+          '2': 'name': 'markup.raw.inline.asciidoc'
       }
 
       # Matches HTML entities


grammars/asciidoc.cson
```diff
@@ -381,8 +381,9 @@
       #
       #   `Text in {backticks}` renders exactly as entered, in `monospace`.
       {
-        'match': '(?<=^|\\s)`[^`]+`'
-        'name': 'markup.raw.inline.asciidoc'
+        'match': '(^|[^\\p{Word};:"\'`\\}])((?:\\[([^\\\\]]+?)\\])?`(\\S|\\S.*?\\S)`(?![\\p{Word}"\'`]))'
```
mojavelinux: Now that we know we can use regexp groups like `\p{Word}`, we should probably start using them more consistently throughout the grammar. The benefit of these groups is that they are friendly to all languages (not just English).

ldez:
If you don't have any remarks, is it possible to merge and the #58 ?
Like that I could begin to create the generator.

mojavelinux:
:+1:

nicorikken:
Did a couple of practical tests and merged. Nice refactor @ldez !

PR #61, opened 2016-04-25
Grammar generator

ldez:
- [x] merge PR #58 and #60 
- [x] split the grammars file into multiple files (partials grammars)
- [x] create a master grammar
- [x] create a grammars generator (use master grammar and partials grammars)
- [x] create a grammars generator for the code blocks (Asciidoctor and Markdown) 
- [x] Document generation system. (and GitHub PR template)

I've created a package for gitignore's like file with a small generator (https://github.com/ldez/atom-language-ignore).
I'm going to use this as starting point.

PR #67, opened 2016-04-28
resolves #65 properly match and label attribute entries

### LINKED ISSUE #65: Attributes whose values contain spaces are not highlighted
E.g.

    :attr: Some text with spaces

According to the documentation, attributes can also contain arbitrary inline content, which is also not highlighted as the complete attribute value is a single colour.

mojavelinux:
- permit any characters in value
- permit name to begin or end with !, but not both
- use more semantic labels
- label punctuation surrounding attribute entry

### DIFF
--- a/grammars/asciidoc.cson
+++ b/grammars/asciidoc.cson
@@ -60,11 +60,15 @@ patterns: [
   #
   #   :icons: font
   #   :sectanchors:
+  #   :!compat-mode:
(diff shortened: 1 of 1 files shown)


nicorikken:
As this was incorporated in issue #61 and that issue was just closed, this PR is closed as well. Thanks for the input :+1:

PR #70, opened 2016-04-29
Adding information about new file extensions.

bwklein:
# Pull Request

I use the .txt extension for my asciidoc files.  While it is not ideal I need this for my current set of tools to work.  This edit to the README file describes the steps to follow to add a new file extension to be recognized by this package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@ Adds syntax highlighting and snippets to AsciiDoc files. Supports [Asciidoctor](
 
 ![](https://raw.github.com/wiki/asciidoctor/atom-language-asciidoc/writers-guide-screenshot.png)
 
+The default file extensions for AsciiDoc files are _ad_, _asc_, _adoc_ and _asciidoc_.  To add a different file extension such as _.txt_, click on the *Settings* button in the package listing and then click the *View Code* button on the settings page.  Find the file named 'language-asciidoc.cson' under 'language-asciidoc > grammars'.  In the list of fileTypes, add a new line below "asciidoc" and type "txt". Then save the file and restart atom or press ctrl+alt+r to refresh the UI.  You should now see the new file type recognized by the atom-language-asciidoc package.
+
 ## Contributing
 
 In the spirit of free software, _everyone_ is encouraged to help improve this project.


nicorikken:
This issue was discussed briefly in Gitter.
**Issue:**
- Some Asciidoc tooling is still based on the .txt file format, so clearly just for backwards compatibility.

**Alternatives offered:**
- Expose the .txt extension as an option on the settings-page
- Support .adoc.txt for .txt compatibility.

mojavelinux:
:+1: 

I also think we should drop `.ad` and `.asc` from the default list. We officially only support `.adoc` and `.asciidoc` in the Asciidoctor toolchain. `.adoc` is preferred. `.asciidoc` must be supported as it's still in wide use, including by the Atom Flight Manual.

I'm very open to supporting `.adoc.txt` by default.

bwklein:
I would be glad to switch to .adoc.txt for my files.  Then when the translation tool supports the adoc file extension and format natively I can drop the .txt part and live happily ever after.

ldez:
I think you can open an issue to discuss on that.

nicorikken:
If indeed we want to reduce the number of supported file-extensions, this requires some further changes. Some of which in the grammar definitions, some in the documentation. I'd like to hint users to change to .asciidoc.txt file extensions, in favor of modifying source-code to hack around the issue. As @bwklein suggests that that would fork for him as well, I think that should be the default route. Should we even document the workaround then?

Either way this PR needs some changes to match the new insights.

nicorikken:
Any comments on:
1. documenting the manual-edit workaround
2. reducing filetype support

or @ldez do you want to move this to tho the issues?

bwklein:
I would say that having a configuration file or a place to specify a list of extensions that the package will recognize and use the asciidoc grammar for would be great.  Other than that, hard coding the list is fine and I would suggest a single combination like '*.adoc.txt' or'*.asciidoc.txt' and then document that for use cases like I have.

ldez:
You can already add custom file/languages associations:

- Menu > `Edit` > `Config...`
- or edit `~/.atom/config.cson`

```coffee
  core:
    ...
    customFileTypes:
      "source.asciidoc": [
          "foo" # all files with `.foo` extension (ex: `documentation.foo`)
          "foobar.txt" # all files with `.foobar.txt` extension  (ex `documentation.foobar.txt`)
      ]
    ...
```

ldez:
This PR can be merge as this.

We edit the file before to add `.adoc.txt` extension.

If you want discuss on remove/add extention please open an issue.

nicorikken:
So to me it seems documenting the currently supported filetypes (not adding or removing any), and writing some guidance for the customFileTypes property would be the consensus solution.

nicorikken:
I rewrote this PR in #89

PR #72, opened 2016-04-29
Use `toHaveLength` instead of `.length).toEqual`

ldez:
Homogenization of style:
- grammars attribute `name` in first place
- Use `toHaveLength` instead of `.length).toEqual`
- ordering capture grammar: `begin` > `pattern` > `end`


nicorikken:
Nice!

ldez:
close in profit of #74

PR #78, opened 2016-05-02
Admonition multi-lines support

ldez:
## Description

Admonition multi-lines support.

## Syntax example

```asciidoc
foobar

NOTE: An admonition paragraph draws the reader's attention to
auxiliary information.
Its purpose is *determined* by the label
at the _beginning_ of the paragraph.

Here are the other built-in admonition types:

TIP: a *pro* _tip_...

IMPORTANT: Don't *forget* icon:megaphone[]

WARNING: Watch out for...

CAUTION: Ensure that...
```

## Screenshots

capture du 2016-05-03 01-12-49

### DIFF
--- a/grammars/language-asciidoc.cson
+++ b/grammars/language-asciidoc.cson
@@ -547,13 +547,13 @@ repository:
         begin: "^((NOTE|TIP|IMPORTANT|WARNING|CAUTION)\\:)\\p{Blank}+"
         captures:
           "1":
-            name: "support.constant.asciidoc"
+            name: "support.constant.label.asciidoc"
         patterns: [
           {
             include: "#inlines"
           }
         ]
-        end: "$"
+        end: "^\\p{Blank}*$"
       }
     ]
   "example-block":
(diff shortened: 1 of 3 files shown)


nicorikken:
Found a bug in your implementation, a newline with just spaces should also end the block:
**source**
screen shot 2016-05-03 at 08 12 48
**render**
screen shot 2016-05-03 at 08 14 36

ldez:
It's fixed ! :smile:

nicorikken:
Nice (funny though that I'm now able to spot implementation errors just by looking at the regex's 😄 )

nicorikken:
Good correction, merged!

mojavelinux:
Nice work!!

My only suggestion is that we should try to use a non-capturing group unless we intend to use it to assign a name or to feed it to a nested pattern.

Here's where this would apply in this case:

```
  begin: '^((?:NOTE|TIP|IMPORTANT|WARNING|CAUTION)\\:)\\p{Blank}+'
```

Notice the addition of `?:`. It's not a big deal, but I think it makes the intent of the patterns clearer. If we capture a group, but don't do anything with it, it seems like maybe we left something out.

mojavelinux:
Technically, a delimited block can terminate a paragraph. However, this is generally considered bad practice and blank lines are preferred. Therefore, I think we should stick with the assumption in this grammar that a paragraph is terminated by a blank line. :+1:

mojavelinux:
Maybe this is the secret to multi-line support for formatted text. We look for either our ending mark or a blank line, whichever comes first.

PR #79, opened 2016-05-02
Add line break support

### LINKED ISSUE #42: A plus sign that indicates continuation should be highlighted
If I have a list and I want a continuation, I use a plus ("+") to do that. However, the plus is not highlighted and looks like plain text.

ldez:
## Description

Add line break support.

## Syntax example

```asciidoc
 +
Foo +

Rubies are *red* +
_Topazes_ are blue.

[%hardbreaks]
Ruby is red.
Java is black.
```

## Screenshots

capture du 2016-05-03 01-15-03

## Issues

Fix #42

### DIFF
--- a/grammars/language-asciidoc.cson
+++ b/grammars/language-asciidoc.cson
@@ -42,6 +42,9 @@ patterns: [
   {
     include: "#inlines"
   }
+  {
+    include: "#line-break"
+  }
 ]
 repository:
   blocks:
@@ -229,6 +232,15 @@ repository:
       {
         name: "constant.other.symbol.horizontal-rule.asciidoc"
         match: "^(?:'|<){3,}$|^ {0,3}([-\\*'])( *)\\1\\2\\1$"
+      }
+    ]
+  "line-break":
+    patterns: [
+      {
+        match: "\\p{Blank}(\\+)$"
+        captures:
+          "1":
+            name: "variable.line-break.asciidoc"
       }
     ]
   list:
--- a/grammars/repositories/asciidoc-grammar.cson
+++ b/grammars/repositories/asciidoc-grammar.cson
@@ -27,6 +27,8 @@ patterns: [
   include: '#blocks'
 ,
   include: '#inlines'
+,
+  include: '#line-break'
 ]
 repository:
   blocks:
--- a/grammars/repositories/partials/line-break-grammar.cson
+++ b/grammars/repositories/partials/line-break-grammar.cson
@@ -1,0 +1,16 @@
+key: 'line-break'
+
+patterns: [
+
+  # Matches a trailing + preceded by at least one space character,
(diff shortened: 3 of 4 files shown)


mojavelinux:
Nice work!

I don't think we should match the hardbreaks options explicitly though. That's one of several ways to set the option and shouldn't get special highlighting unique from other options. I think we should just focus on the explicit line break character.

nicorikken:
Nice work indeed. I actually had a colleague working on this issue to get him into Asciidoctor development 😝 He was going for an intelligent multi-line solution, but this PR is more practical (as we now know multi-lines an be a pain in the #$%@).
Regarding the [%hardbreaks], I personally prefer highlighting on valid keywords, to have confidence on the validity when writing. But I do recognize this option could be integrated with more generic grammar.

mojavelinux:
>  I personally prefer highlighting on valid keywords

I'm definitely okay with highlighting valid keywords. I just don't think we should be matching the whole block `[%hardbreaks]` as other things should go in there. This is where we should pass the value between the square brackets to an attribute parsing rule to match known keywords.

mojavelinux:
In other words, it's perfectly valid to have:

```
[.poem%hardbreaks]
Strange things did happen here
No stranger would it be
If we met at midnight
In the hanging tree.
```

ldez:
I'm going to remove `hardbreak`.
I really interesting by known the rules behind this case.

mojavelinux:
The hardbreaks option (written either as `%hardbreaks` or opts=hardbreaks) was added to simplify writing things like addresses. It just saves from having to type the + over and over.

The documentation for the shorthand syntax in the user manual is not very clear. The raw material that is there starts at http://asciidoctor.org/docs/user-manual/#id.

ldez:
@nicorikken others remarks ? If not, can you merge ?
I want to use line break pattern in other patterns.

nicorikken:
Nope, no further comments. Held the merge at the time because of the hardbreak remark. Now it is as discussed. Merged! 👍

mojavelinux:
I don't think we should require there to be at least one non-whitespace character on the line. Although there are cases when this is supported in Asciidoctor, it could be removed in the future.

First, the following is a literal paragraph because it starts with a space:

```
 +
Foo +
Bar
```

If you want a blank line, then you should use the empty attribute.

```
Foo +
{empty} +
Bar
```

Therefore, we should amend the regex to look for at least one non-blank character before the current match.

ldez:
Do something like that `(?:\S)\p{Blank}(\+)$` cause conflicts with others patterns.
My first approach is this, but I failed with this.

mojavelinux:
Shouldn't it be a positive look-behind?

```
(?<=\S)\p{Blank}+\+$
```

mojavelinux:
I'll submit a PR.

PR #81, opened 2016-05-03
Split 'partial/various-grammar.cson' and tests

ldez:
Just split `partial/various-grammar.cson` in multiple grammar files and add more tests.

It's the last splitting action !

Now go to the new features and a surprise...


nicorikken:
I'm a bit confused about the last sentence, but the refactoring seems great, and it includes a large set of additional testcased 👍 Merge!

PR #83, opened 2016-05-07
Add new languages support for code block

ldez:
- Mustache
- OCaml
- Scala
- Groovy
- JSX

### DIFF
--- a/grammars/language-asciidoc.cson
+++ b/grammars/language-asciidoc.cson
@@ -1033,6 +1033,35 @@ repository:
         end: "(?<=----)[\\r\\n]+$"
       }
       {
+        begin: "^\\[source,\\s*(?i:(groovy))\\]$"
+        beginCaptures:
+          "0":
+            name: "support.asciidoc"
+        patterns: [
+          {
+            name: "markup.code.groovy.asciidoc"
+            begin: "^(-{4,})\\s*$"
+            beginCaptures:
+              "0":
+                name: "support.asciidoc"
+            end: "^\\1*$"
+            endCaptures:
+              "0":
+                name: "support.asciidoc"
+            contentName: "source.embedded.groovy"
+            patterns: [
+              {
+                include: "#block-callout"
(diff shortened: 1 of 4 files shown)


mojavelinux:
Groovy!

nicorikken:
As I recently played with Jenkinsfiles, I know the Jenkinsfile syntax is currently best handled by the Groovy language. It is also available as a supported filetype in the Groovy language grammar:

```
'fileTypes': [
  'groovy'
  'gvy'
  'gradle'
  'Jenkinsfile'
]
```

What do you think? Do we need a special language defined, currently handled by the Groovy syntax? (I could imagine such a thing, just like Ansible syntax is a variant of the YAML syntax and up until recently no dedicated Ansible parser was available) or do you want two match-options for the Groovy syntax like `(groovy|jenkinsfile)`?

ldez:
Note if you add `jenkinsfile` to block it's not render properly (in Github but also others rendering tools)

- Groovy

```groovy
class AGroovyBean {
  String color
}

def myGroovyBean = new AGroovyBean()

myGroovyBean.setColor('baby blue')
assert myGroovyBean.getColor() == 'baby blue'

myGroovyBean.color = 'pewter'
assert myGroovyBean.color == 'pewter'
```

- Jenkinsfile

```jenkinsfile
class AGroovyBean {
  String color
}

def myGroovyBean = new AGroovyBean()

myGroovyBean.setColor('baby blue')
assert myGroovyBean.getColor() == 'baby blue'

myGroovyBean.color = 'pewter'
assert myGroovyBean.color == 'pewter'
```

ldez:
Add custom markers, create a gap between editor and rendering tools, I think we must stay the more possible near of rendering tools.

nicorikken:
Agreed, so we give each language a different parser, although sometimes the best parser might be used for some other language as well. Then this one can be merged. Even though I I might add some more language in the future then 😉

PR #84, opened 2016-05-07
correct monospaced phrase matches

mojavelinux:
- don't capture preceding character
- don't match escaped monospaced pattern
- simply match for attribute list
- use non-capturing groups

### DIFF
--- a/grammars/language-asciidoc.cson
+++ b/grammars/language-asciidoc.cson
@@ -434,11 +434,11 @@ repository:
     patterns: [
       {
         name: "markup.raw.constrained.monospaced.asciidoc"
-        match: "(^|[^\\p{Word};:\"'`}])(?:\\[([^\\]]+?)\\])?`(\\S|\\S.*?\\S)`(?![\\p{Word}\"'`])"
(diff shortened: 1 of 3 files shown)


mojavelinux:
Syntax example:

```
(`monospaced`)

\`monospaced
```

Old screenshot:

old

New screenshot:

new

mojavelinux:
There are a few things to keep in mind when looking at the regular expressions in Asciidoctor core (asciidoctor.rb):

1. The regular expressions in core are written for the Ruby 1.8 regexp engine, so they are more primitive than what Oniguruma supports. Most notably, the Ruby 1.8 regexp engine doesn't support look-behind matches (We're going to start using the Oniguruma engine in Asciidoctor 1.6).
2. The regular expressions in core often captures groups in order to populate the AST node, or to perform more fine-grained parsing. The grammar doesn't need to capture a group unless that span of text is needed for highlighting.
3. The grammar should skip matching escaped syntax. Core captures it only because the regexp engine doesn't support look-behind matches.

nicorikken:
@mojavelinux I turned your comment in a PR #86

nicorikken:
Nice fix, tested, so merge!

mojavelinux:
:+1:

PR #89, opened 2016-05-07
doc: file ext support

nicorikken:
To summarize the discussion in PR #70, where the explicit info on the .txt support is desired, but the recommended customization is through the Atom config, rather than through the Grammar modification.

Credits go to @bwklein and @ldez


nicorikken:
Noting the thumbsup of @lddez , the discussion in the previous PR by @bwklein, I merge this PR.

bwklein:
Thank you. :)

PR #90, opened 2016-05-07
make hard line break match more strict

mojavelinux:
Require that the line have at least one non-blank character before the line break syntax.

### DIFF
--- a/spec/partials/line-break-grammar-spec.coffee
+++ b/spec/partials/line-break-grammar-spec.coffee
@@ -18,17 +18,16 @@ describe 'Should tokenize line break when', ->
 
   it 'contains simple character', ->
     tokens = grammar.tokenizeLines '''
-       +
       Foo +
+      Bar
       '''
     expect(tokens).toHaveLength 2
-    expect(tokens[0]).toHaveLength 2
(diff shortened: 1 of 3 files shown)


nicorikken:
Nice addition, avoiding the `+` to be highlighted on an empty line.

spec/partials/line-break-grammar-spec.coffee
```diff
@@ -18,17 +18,16 @@ describe 'Should tokenize line break when', ->
 
   it 'contains simple character', ->
     tokens = grammar.tokenizeLines '''
-       +
       Foo +
+      Bar
```
ldez: My example come from AsciiDoctor code https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor.rb#L970-L983

Are you sure of your change ?

I found a mistake in the AsciiDoctor code comment :
https://github.com/asciidoctor/asciidoctor/blob/master/lib/asciidoctor.rb#L664-L678

Roman letters cannot be follow by a `.` but by a `)` :wink:

mojavelinux: Yep, the sample is asciidoctor.rb is incorrect (in both cases).

While those examples are usually correct, they aren't gospel :)

PR #91, opened 2016-05-08
resolves #43 support text formatting in footnote content

### LINKED ISSUE #43: Italics break when immediately inside footnote
This doesn’t work: `footnote:[_high_ hello]`

Whereas this is fine: `footnote:[hello _high_]`

mojavelinux:
- match inline patterns inside footnote content
- change name of match to markup.other.footnote.asciidoc
- skip escaped footnote
- add CSS for footnote (map to function color)

### DIFF
--- a/styles/asciidoc.atom-text-editor.less
+++ b/styles/asciidoc.atom-text-editor.less
@@ -76,6 +76,10 @@ atom-text-editor, :host {
           font-weight: bold;
       }
 
+      &.footnote {
+          color: @syntax-color-function;
+      }
+
       &.super {
         vertical-align: super;
         font-size: @font-size * .8;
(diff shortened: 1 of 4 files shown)


nicorikken:
I tried this one, works nicely.

styles/asciidoc.atom-text-editor.less
```diff
@@ -76,6 +76,10 @@ atom-text-editor, :host {
           font-weight: bold;
       }
 
+      &.footnote {
```
ldez: same thing as #94 , for now try to avoid custom style, use the 'default' codes create a better cross-theme support.

I note we must make a special effort to refactor styling so to have a more homogeneous style.

nicorikken:
@ldez Do you agree with still merging this in its current state, leaving open that the styles need refactoring in naming/styling?

ldez:
I agree. Merge is ok.

Issue #53, opened 2016-04-16
Update terminology used in grammar file


mojavelinux:
There are a lot of places in the grammar file where the terminology used doesn't match the terminology as defined by AsciiDoc. Update the terminology in the grammar file so it's consistent with the language definition.

For example, AsciiDoc prefers:

* "strong" instead of "bold"
* "emphasis" instead of "italic"
* "attribute reference" instead of "substitution"
* "xref" instead of "reference"
* "monospaced" instead of "raw"
* "char(acter) reference" (or "entity") instead of "htmlentity"
* "url" instead of "link" (and certainly not "underline")

Additionally, we shouldn't be matching explicit paragraph block style names on top of paragraphs as those names are extensible.

nicorikken:
Although I agree with this proposed change regarding the naming of partial grammar, and custom style elements, if I'm not mistaken (did a small test to check), deviating from the commonly used descriptors like bold, italic, can break current theme compatibility. That seems less desirable to me, as it might take a while for themes to catch up. What do you think?

mojavelinux:
I'm okay with using the common terminology in the names. Where I'd like to see the update is in the comments and in the file names. In other words, we should describe what we are matching using the AsciiDoc syntax, but use the closest terminology known to themes for the assigned names.

For example, "Matches bold unconstrained phrases" should be "Matches strong unconstrained phrases".

We could also consider adding both the AsciiDoc terminology in the name and the closest match to the terminology in the themes. For example: "markup.bold.strong.unconstrained.asciidoc".

(Btw, I don't see much value in using "constrained" and "unconstrained" in the name value. I don't see any case where these would be highlighted differently).

ldez:
"constrained" and "unconstrained" in the name value is very useful to debug. :wink:

mojavelinux:
Hmm, I see.

nicorikken:
@ldez I removed the 'constrained' 'unconstrained' distinction in the PR #94 For debugging it is pretty simple to disable either of the two, or to rename them. So although I see the debugging use-case, I opt for simplicity. What do you think?

mojavelinux:
We have an inconsistency in how we refer to formatted text (which I agree, is true in core as well).

* strong
* emphasis
* monospaced

I think it should either be:

* strong
* emphasized
* monospaced

or

* strong
* emphasis
* monospace

I prefer the second. So we should drop the "d" in "monospaced".

ldez:
[...]

PR #94, opened 2016-05-08
feat: meta.attrlist style for strong

nicorikken:
This feature adds the markup.meta.attrlist.asciidoc attribute for strong content, matching the whole attribute list, including the surrounding block quotes.

This addresses issue #9

Also the "bold" type has been refactored to read "strong", to address issue #53

Before:
screen shot 2016-05-08 at 11 56 27
After:
screen shot 2016-05-08 at 11 56 08


styles/asciidoc.atom-text-editor.less
```diff
@@ -87,6 +87,10 @@ atom-text-editor, :host {
         font-size: @font-size * .8;
         color: mix(green, @syntax-text-color, 50%);
       }
+
+      &.meta.attrlist {
```
ldez: Try to avoid custom style, use the 'default' code create a better cross-theme support.

I will try you're PR to see the rendering.

ldez:
Can you provide a syntax example, so I could quickly see the change?

styles/asciidoc.atom-text-editor.less
ldez: I have try and I'm ok with 'custom' styling.

Can you simple adding `&.meta.attrlist` to the existing  style section like that ?

```less

&.heading.blocktitle,
      &.substitution,
      &.meta.attrlist {
        color: @syntax-text-color-unobtrusive;
      }
```

nicorikken:
So, mergeable?

ldez:
Once you have changed the style sheet :wink:

ldez:
You must rebase. sorry it's my fault.

nicorikken:
Not your fault, right order, I'll correct and probably add attribute lists for other styles.

grammars/language-asciidoc.cson
```diff
@@ -587,18 +587,22 @@ repository:
   strong:
     patterns: [
       {
-        name: "markup.bold.unconstrained.asciidoc"
-        match: "\\\\?(?:\\[([^\\]]+?)])?(\\*\\*)(.+?)(\\*\\*)"
+        name: "markup.bold.strong.asciidoc"
```
ldez: I see your comment on https://github.com/asciidoctor/atom-language-asciidoc/issues/53#issuecomment-217783602

I prefer keep `unconstrained` and `constrained` because if I disable a rule, I don't show properly the edge effects.

mojavelinux:
After studying http://manual.macromates.com/en/language_grammars.html#naming_conventions more, I don't think "markup.meta.attrlist.asciidoc" is the right choice for name.

Here are my alternate proposals, in order of preference:

* markup.other.attrlist.asciidoc
* support.other.attrlist.asciidoc

wdyt? Or do you think we do have it right with "markup.meta.attrlist.asciidoc"?

nicorikken:
Reading the naming conventions, and thinking long and hard about it (rather vague and seemingly overlapping definitions) I'd opt for 'markup.meta.attrlist.asciidoc' as it is bound to a markup language, but is not related to a particular markup style, it is in that sense 'a larger construct' of attribute entities relating to the markup. 'markup.other.attrlist.asciidoc' is a 2nd-best to me. 'support.other.attrlist.asciidoc' seems to convey predefined attributes, which can be the case, but isn't required. Can you agree with 'markup.meta.attrlist.asciidoc' ?

mojavelinux:
I like your reasoning to stick with "markup.meta.attrlist.asciidoc". If we adopt that, we should use the "markup.meta" prefix anytime we are highlighting something that "attaches to" content that is being marked up.

The AsciiDoc bundle for TextMate uses "support.variable.attributelist.asciidoc", so another option to consider. I prefer "attrlist" (or "attribute-list") over "attributelist", so we'd use "support.variable.attrlist.asciidoc" (or "support.variable.attribute-list.asciidoc").

See https://github.com/mattneub/AsciiDoc-TextMate-2.tmbundle/blob/master/Syntaxes/asciidoc.tmLanguage (nice reference in general, btw).

> rather vague and seemingly overlapping definitions

exactly my impressions

nicorikken:
The current state:
screen shot 2016-05-10 at 07 35 18
What do you think about the readability of the attribute-list? In this way the context is clearly visible based on the style, but readability seems to be taking a hit.

ldez:
Why tags characters have this color ? Why sub and sup applying on role attribute ?

nicorikken:
I duplicated the pattern used for strong content, using 'support.constant.asciidoc' for the markers. Why any distinction?

mojavelinux:
I don't think the formatting marks should be colored differently than the formatted text. Perhaps we should switch to using something that doesn't get a style by default. The gfm and rst grammars use `punctuation.definition...`. Perhaps we should use `punctuation.definition.asciidoc`. I don't think it is right to say that they are constants.

mojavelinux:
Another alternative is `markup.other.punctuation`.

mojavelinux:
I think that the attribute list should be displayed consistently. In other words, always:

* gray
* normal weight
* normal style

Obviously, superscript and subscript would be an exception.

mojavelinux:
Aside from that, it looks amazing! I can't wait for this. We're really pushing ahead of other grammar implementations with this one!

nicorikken:
I also saw `punctuation.definition.entity.gfm` definition listed in the gfm grammar. `markup.other.punctuation.asciidoc` seems to make more sense as the TextMate documentation doesn't list `punctuation` as a main classification. Conforming to other Atom packages seems to be the winning argument for me, so I'll adopt `punctuation.definition.entity.gfm` then.

I would need to override the markup style (`strong`, `emphasis`) to avoid italic or bold fonts being rendered.

@ldez Why would you prefer different colors for sub and sup? I would expect a similar behavior as the other styles, so no color distinction.

nicorikken:
Current state:
screen shot 2016-05-10 at 18 49 44

grammars/repositories/inlines/emphasis-grammar.cson
```diff
@@ -2,23 +2,29 @@ key: 'emphasis'
...
+  # Matches emphasis unconstrained phrases
   #
   # Examples
   #
-  #   _italic phrase_
+  #   e__mphasis__ phrase
   #
-  match: '(?!_{4,}\\s*$)(?<=^|[^\\w;:])(?:\\[(?:[^\\]]+?)\\])?(_\\S_|_\\S[\\s\\S]*?\\S_)(?!\\w)'
+  name: 'markup.italic.emphasis.unconstrained.asciidoc'
+  match: '(?<!\\\\\\\\)(\\[(?:[^\\]]+?)\\])?(__)(?:(?!_).+?)(__)' # now including the underscores in the match
```
ldez: move the inline comment to beginning of the previous line

styles/asciidoc.atom-text-editor.less
```diff
@@ -59,7 +59,12 @@ atom-text-editor, :host {
       }
 
       &.heading.blocktitle,
-      &.substitution {
+      &.substitution{
+        color: @syntax-text-color-unobtrusive;
+      }
+      &.meta.attrlist {
+        font-weight: normal;
```
ldez: Styling bad practice.
I see why you do that but you must change the captures group instead of doing that.

ex:

```coffee
  name: 'markup.emphasis.unconstrained.asciidoc'
  # now including the underscores in the match
  match: '(?<!\\\\\\\\)(\\[(?:[^\\]]+?)\\])?((__)((?!_).+?)(__))'
  captures:
    1: name: 'markup.meta.attrlist.asciidoc'
    2: name: 'markup.italic.emphasis.asciidoc'
    3: name: 'punctuation.definition.entity.asciidoc'
    4: name: 'punctuation.definition.entity.asciidoc'
```

styles/asciidoc.atom-text-editor.less
```diff
@@ -59,7 +59,12 @@ atom-text-editor, :host {
...
+        font-style: normal;
```
ldez: Styling bad practice.
I see why you do that but you must change the captures group instead of doing that.

ex:

```coffee
  name: 'markup.emphasis.unconstrained.asciidoc'
  # now including the underscores in the match
  match: '(?<!\\\\\\\\)(\\[(?:[^\\]]+?)\\])?((__)((?!_).+?)(__))'
  captures:
    1: name: 'markup.meta.attrlist.asciidoc'
    2: name: 'markup.italic.emphasis.asciidoc'
    3: name: 'punctuation.definition.entity.asciidoc'
    4: name: 'punctuation.definition.entity.asciidoc'
```

nicorikken: I was already thinking the same thing, but the current fix was quick and dirty. :wink:  I will improve it.

ldez:
also change `monospaced-grammar.cson` to `monospace-grammar.cson`

ldez:
Current rendering state is nice for me.

ldez:
I found regressions:

capture du 2016-05-10 22-24-03
capture du 2016-05-10 22-24-51


```adoc
image::supporting-image.png[]

image:supporting-image.png[]
icon:download[link="http://rubygems.org/downloads/asciidoctor-1.5.2.gem"]

stem:[x != 0]
asciimath:[x != 0]
asciimath:dsqdsq[x != 0]
latexmath:[\sqrt{4} = 2]
```

mojavelinux:
That output is looking nice.

I hate to do this to you, but I think I prefer `punctuation.definition.<type>.asciidoc` (where `<type>` is bold, italic, raw, etc). That's more consistent with the textmate bundle for Markdown. I don't think "entity" really fits here.

I also now prefer the longer form `markup.meta.attribute-list.asciidoc` for the attribute list.

nicorikken:
Current rendering
screen shot 2016-05-11 at 07 35 31

nicorikken:
Current rendering:
screen shot 2016-05-11 at 07 58 44

mojavelinux:
Looking good! A few things:

* We should get this in the 1.2.0 release
* The formatting marks should either match the color of the text or be
slightly more transparent (which would actually be rather cool)
* Superscript & subscript don't need to be a different color since they
already look distinct

ldez:
I prefer sub and sup with color, can we stay with current colors state ?

nicorikken:
I was trying to get the matching colors on the punctuation to work, but I need to read through some introduction material on the LESS paradigm to improve this. Maybe @ldez you can help? Otherwise we can still stick with the current colors for 1.2.0?

ldez:
Merge the current state and I will do a styling PR if you want ?

nicorikken:
Please, I'll merge this one then.

mojavelinux:
I think we're almost there with the naming, but there's still some refinement needed.

I don't think we need to wrap the formatted text itself. It's sufficient that it be inherited from outer pattern.

Here's what I recommend:

```
markup.strong.unconstrained.asciidoc - [role]**strong text**
  meta.attribute-list.asciidoc - [role]
  punctuation.definition.uncontrained.asciidoc - **
  (no name) - strong text
  punctuation.definition.uncontrained.asciidoc - **
```

I think we can rely more an CSS inheritance to keep the names simpler. I repeat the "unconstrained" because it qualifies the "punctuation.definition". Another option is "markup.punctuation.uncontrained".

wdyt? Follow-up issue?

I think it's really important to get these names right because a lot will depend on these names downstream.

mojavelinux:
I think the problem you are having the punctuation is the addition of `.bold` in the name. I think if we drop that part, it should be easier to work with the styles. Having given it more thought, I don't see why we need to repeat that on the punctuation itself.

mojavelinux:
I'm pretty happy with the naming looking at my last proposal.

ldez:
Open a issue for discuss on this.

nicorikken:
I just created an issue #110 to continue this discussion.

PR #100, opened 2016-05-08
Email highlight

ldez:
Support for email highlight.

```adoc
foo@bar.com
```

Regex from AsciiDoctor code.


nicorikken:
Code reviewed, and tested. Seems based on `EmailInlineMacroRx` from the upstream Asciidoctor.rb project.
screen shot 2016-05-09 at 08 07 06

PR #101, opened 2016-05-08
Rename language-properties.cson

ldez:
Rename `properties-lang.cson` to `language-properties.cson`


ldez:
There is no impact from this change. I merge.

PR #107, opened 2016-05-10
feat: use Shadow DOM

ldez:
> the shadow DOM is now enabled by default to prevent style pollution in Atom’s text editor component.

http://blog.atom.io/2015/01/05/enable-shadow-dom.html


nicorikken:
If I understand correctly, this prevents styling conflicts with other packages, because the style is now sort of 'contained' in its own root. Seems like a good practice. Is there any way I can test this before the merge? The code itself seems simple enough.

ldez:
There's no easy way to test this.

nicorikken:
Alright, I'll merge this then, based on unit tests and associated documentation.

PR #116, opened 2016-05-12
feat: multi-lines

ldez:
Basic multi-lines for unambiguous elements.

### DIFF
--- a/grammars/language-asciidoc.cson
+++ b/grammars/language-asciidoc.cson
@@ -620,13 +620,13 @@ repository:
   "stem-macro":
     patterns: [
       {
-        name: "markup.macro.inline.stem.general.asciidoc"
-        match: "(?<!\\\\)(stem|(?:latex|ascii)math):([a-z,]*)(\\[)(.*?[^\\\\])(\\])"
-        captures:
+        name: "markup.macro.inline.stem.asciidoc"
+        begin: "(?<!\\\\)(stem|(?:latex|ascii)math):([a-z,]*)(\\[)"
+        beginCaptures:
           "2":
-            name: "support.constant.asciidoc"
-          "4":
-            name: "support.constant.asciidoc"
+            name: "markup.meta.attribute-list.asciidoc"
+        contentName: "markup.raw.asciidoc"
+        end: "\\]|^$"
       }
     ]
   strong:
--- a/grammars/language-asciidoc.cson
+++ b/grammars/language-asciidoc.cson
@@ -694,21 +694,25 @@ repository:
     patterns: [
       {
         name: "markup.reference.xref.asciidoc"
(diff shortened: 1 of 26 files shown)


ldez:
@nicorikken @mojavelinux Is that kind of behavior acceptable to you?

```adoc
foobar

footnoteref:[id, text text
text text text]

foobar

footnoteref:[id, text text
text text
text

foobar
```

capture du 2016-05-12 19-56-31

grammars/language-asciidoc.cson
```diff
@@ -471,17 +471,17 @@ repository:
     patterns: [
       {
         name: "markup.other.footnote.asciidoc"
-        match: "(?<!\\\\)(footnote(?:ref)?):\\[(.*?[^\\\\])\\]"
```
nicorikken: Why did you exchange the lookbehind for a match?

nicorikken:
Somewhat better, assuming people don't make mistakes like leaving out the closing brackets. But that would better be handled using a linter.
Some of my edge case testing:
screen shot 2016-05-12 at 20 04 15
```asciidoc
foobar

footnoteref:[id, text text
text text text]

foobar

footnoteref:[id, text text
text text
text

foobar
```

I think this is still better than the single-line variant, assuming valid Asciidoctor syntax.

grammars/language-asciidoc.cson
ldez: bad copy paste. fix

ldez:
If you assuming this is possible, I apply this to:
- footnote(ref)
- passthrough (only the `pass:[text]` variant)
- stem
- xref

nicorikken:
Certainly I think you can add this to the other constructs with a similar (bounded) syntax. That leave only the more conflicting constructs up for debate.

ldez:
I can do that:

capture du 2016-05-12 21-16-12

Is this expected behavior?

ldez:
It's ready to review.

Say me if spaces at the ending of line are invalid characters.

ldez:
I have also fixed all macros for respect naming conventions and styling.

nicorikken:
I see no breakage in practical coloring (due to renaming). Code is nice. Spaces at the end seem to render neatly in Asciidoc preview plugin, so would be valid. As long as the macro is closed.

This can be merged.

mojavelinux:
> Say me if spaces at the ending of line are invalid characters.

The first thing the converter does is strip all trailing whitespace. So technically it is valid because it is ignored. But we could highlight it as an error to show that it will get stripped away and is thus superfluous.

mojavelinux:
> Somewhat better, assuming people don't make mistakes like leaving out the closing brackets. But that would better be handled using a linter.

True. It's also unavoidable given the current behavior of first-mate. They really, really, really need to add a way to validate the inner content against a spec. Right now, this is the very best we can do.

PR #117, opened 2016-05-12
feat: checklist snippets

nicorikken:
Two snippets for the easy creation of both unchecked and checked checklist items.
Originally I was aiming for the `-[` snippet, but that raises issues if the editor automatically closes the first `[` bracket with a second `]` bracket after the cursor. Using the lower-case characters `o` and `x` resolve this, whilst symbolizing the check rendering.
screen shot 2016-05-12 at 19 56 08


ldez:
LGTM

nicorikken:
I've now been using this daily since I pushed it, works great for my use-case, nice and easy. Thanks for the merge.

PR #122, opened 2016-05-14
feat: url with attributes references

ldez:
## Description

Use attributes references with url

## Syntax example

```adoc
{uri-repo}[Asciidoctor]
[{uri-repo}[Asciidoctor]
({uri-repo}[Asciidoctor]
<{uri-repo}[Asciidoctor]
>{uri-repo}[Asciidoctor]
foo {uri-repo}[Asciidoctor] bar

link:http://{uri-repo}[label]
[link:{uri-repo}[label]
(link:{uri-repo}[label]
<link:{uri-repo}[label]
>link:{uri-repo}[label]
foo link:{uri-repo}[label] bar
Go on link:https://foobar.com now

http://foo{uri-repo}bar.com
<http://foobar.com[foobar]
>http://foobar.com[foobar]
(http://foobar.com[foobar]
[http://foobar.com[foobar]
foo http://foobar.com[foobar] bar
```

## Screenshots

capture du 2016-05-14 17-26-19

### DIFF
--- a/grammars/language-asciidoc.cson
+++ b/grammars/language-asciidoc.cson
@@ -150,7 +150,41 @@ repository:
     patterns: [
       {
         name: "meta.definition.attribute-entry.asciidoc"
-        match: "^(:)(!?\\w[\\p{Blank}\\w-]*?|\\w[\\p{Blank}\\w-]*?!?)(:)(\\p{Blank}+.*)?$"
+        begin: "^(:)(!?\\w.*?)(:)(\\p{Blank}+.+\\p{Blank}\\+)$"
+        beginCaptures:
+          "1":
+            name: "punctuation.separator.attribute-entry.asciidoc"
+          "2":
+            name: "support.constant.attribute-name.asciidoc"
+          "3":
+            name: "punctuation.separator.attribute-entry.asciidoc"
+          "4":
+            name: "string.unquoted.attribute-value.asciidoc"
+            patterns: [
+              {
+                include: "#inlines"
+              }
+              {
+                include: "#line-break"
+              }
+            ]
+        contentName: "string.unquoted.attribute-value.asciidoc"
+        patterns: [
+          {
+            include: "#inlines"
+          }
+          {
+            include: "#line-break"
+          }
+        ]
+        end: "^\\p{Blank}+.+$(?<!\\+)|^\\p{Blank}*$"
+        endCaptures:
+          "0":
+            name: "string.unquoted.attribute-value.asciidoc"
+      }
+      {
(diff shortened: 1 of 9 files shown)


mojavelinux:
This is a great example of where the grammar sees something that the parser never does. By the time the parser sees the link, the attribute has been resolved. But to the writer, the attribute is part of the URL.

I'd like to propose a slight modification. I'd like to see the `uri-` prefix required in order to make this match (e.g., `{uri-google}[Google]`). That does two things. First, it promotes the best practice of adding the `uri-` prefix to the name of a URI attribute. Second, it cuts out the ambiguity of matching the pattern `{foo}[Bar]` when it is not a link.

mojavelinux:
When the explicit `link:` prefix is specified, then we can assume it is always a link (e.g., `link:{foo}[Bar]`).

ldez:
Done!

mojavelinux:
I love it!

ldez:
small bonus: attribute multi-lines who respect the line-break :wink: 

capture du 2016-05-16 23-52-39

mojavelinux:
Sweet like :candy: !

mojavelinux:
> small bonus: attribute multi-lines who respect the line-break

...actually, I almost missed it. In Asciidoctor, we now prefer a backslash (`\`) instead of a plus (`+`) at the end of the line (like Bash). Could you switch it?

Asciidoctor also supports a hardbreak by using `+ \` at the end of the line. Example:

```
:address: Street + \
City, State ZIP

{address}
```

ldez:
Supports backslash line-break and backslash hard-break.

capture du 2016-05-19 19-47-41

mojavelinux:
:+1:

nicorikken:
So @ldez if I understand it correctly, your wizardry fix is to have a very generic begin-end match, and only include line breaks with continuation characters, next to the inline styles? Very clever. This will be something to watch out for when we work on multi-line inlines like bold and italics ⚠️ because in paragraphs line-continuation will be handled differently:
```
:description: test *failing
multi-line*
```
```
:description: test *succeeding \
multi-line*
```

I guess we're going to need a separate library (include) for multi-line inline styles, to enable this type of line continuation in other parts.

Oh and by the way 😉 nice work on the URL matching. I assume you are going to pick up on the `uri-` suggestion?

ldez:
The `uri-` suggestion is already done :wink:

nicorikken:
My bad, I assumed it was emphasised in highlighting. Reading back I now see it was about restricting the regex. Alright then. LGTM

PR #124, opened 2016-05-14
feat: definition list item / labeled list

ldez:
## Description

Support for definition list item / labeled list

## Syntax example

```adoc
CPU:: The brain of the computer.
Hard drive:: Permanent storage for operating system and/or user files.
RAM:: Temporarily stores information the CPU uses during operation.
Keyboard:: Used to enter text or control items on the screen.
Mouse:: Used to point to and select items on your computer screen.
Monitor:: Displays information in visual form using text and graphics.
```
```adoc
Operating Systems::
  Linux:::
    . Fedora
      * Desktop
    . Ubuntu
      * Desktop
      * Server
  BSD:::
    . FreeBSD
    . NetBSD

Cloud Providers::
  PaaS:::
    . OpenShift
    . CloudBees
  IaaS:::
    . Amazon EC2
    . Rackspace
```

## Screenshots

capture du 2016-05-14 20-05-45


mojavelinux:
I like it!

mojavelinux:
Technically, there can be a space before the delimiter.

```
CPU :: The brain of the computer.
```

However, I'm glad you are not supporting it because we are discussing disallowing the space in https://github.com/asciidoctor/asciidoctor/issues/1066.

nicorikken:
Very nice. Regarding the space, either solution would be fine to me, especially since the current user guide doesn't show the use of a preceding space. I think this can be merged.

PR #128, opened 2016-05-16
feat: add mark support

ldez:
## Description

http://asciidoctor.org/docs/user-manual/#custom-styling-with-attributes

## Syntax example

```adoc
.foobar
[role]**foo** bar
[role]*foo* bar
[role]__foo__ bar
[role]_foo_ bar
[role]##foo## bar
[role role2]#foo# bar
[role]``foo`` bar
[role]`foo` bar
[role]^foo^bar^foo^bar
[role]~foo~bar~foo~bar
```

## Screenshots

capture du 2016-05-16 21-55-19

### DIFF
--- a/grammars/language-asciidoc.cson
+++ b/grammars/language-asciidoc.cson
@@ -709,12 +709,36 @@ repository:
     patterns: [
       {
         name: "markup.mark.unconstrained.asciidoc"
-        match: "(?<!\\\\\\\\)(\\[[^\\]]+?\\])?((##)(.+?)(##))"
+        match: "(?<!\\\\\\\\)(\\[[^\\]]+?\\])((##)(.+?)(##))"
         captures:
           "1":
             name: "markup.meta.attribute-list.asciidoc"
           "2":
+            name: "markup.mark.asciidoc"
+          "3":
+            name: "punctuation.definition.asciidoc"
+          "5":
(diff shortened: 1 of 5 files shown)


mojavelinux:
Nice job!

Mark without a role becomes the mark element in HTML, which appears as highlighted text. Would it be possible to style it that way in the theme? (basically by reversing the background and foreground colors).

ldez:
Like that ? :smile: 

capture du 2016-05-16 23-44-49

```adoc
[role role2]##foo## bar
[.role#id]##foo## bar
##foo## bar
#foo# bar
```

mojavelinux:
Yes! \o/

mojavelinux:
I think we're in good shape. @nicorikken, what say you?

nicorikken:
Somehow I haven't been able to get the highlight to work under my theme. But syntax highlighting looks good. LGTM!

ldez:
I have try with different themes, which theme are you using ?

nicorikken:
I tried Seti, Atom One Dark, Atom One Light.

ldez:
one-dark:
one-dark

one-light:
one-light

seti:
seti

nicorikken:
Apparently I did not check with the latest code. Now it does work properly 👍

PR #129, opened 2016-05-18
feat: typographic quotes

ldez:
## Description

Quotation Marks and Apostrophes: http://asciidoctor.org/docs/user-manual/#curved

## Syntax example

```adoc
foo [bar]"`double-quoted`" foo
foo [bar]'`single-quoted`' foo
foo "`double-quoted`" foo
foo '`single-quoted`' foo
```

## Screenshots

capture du 2016-05-18 21-15-57

### DIFF
--- a/grammars/language-asciidoc.cson
+++ b/grammars/language-asciidoc.cson
@@ -863,28 +863,28 @@ repository:
   "typographic-quotes":
     patterns: [
       {
-        name: "markup.other.typographic-quotes.asciidoc"
+        name: "markup.quote.typographic-quotes.asciidoc"
         comment: "double-quoted"
         match: "(?:^|(?<!\\p{Word}|;|:))(\\[([^\\]]+?)\\])?(\"`)(\\S|\\S.*?\\S)(`\")(?!\\p{Word})"
         captures:
           "1":
             name: "markup.meta.attribute-list.asciidoc"
           "3":
-            name: "constant.other.symbol.asciidoc"
+            name: "punctuation.definition.asciidoc"
           "5":
-            name: "constant.other.symbol.asciidoc"
+            name: "punctuation.definition.asciidoc"
       }
       {
-        name: "markup.other.typographic-quotes.asciidoc"
+        name: "markup.quote.typographic-quotes.asciidoc"
         comment: "single-quoted"
         match: "(?:^|(?<!\\p{Word}|;|:))(\\[([^\\]]+?)\\])?('`)(\\S|\\S.*?\\S)(`')(?!\\p{Word})"
         captures:
           "1":
             name: "markup.meta.attribute-list.asciidoc"
           "3":
-            name: "constant.other.symbol.asciidoc"
+            name: "punctuation.definition.asciidoc"
           "5":
-            name: "constant.other.symbol.asciidoc"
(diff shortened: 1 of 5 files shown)


mojavelinux:
:+1: 

Do you think we should highlight the whole phrase, or would that be distracting? (VIM highlights the whole phrase).

ldez:
only the quotes are affected, the render of the phrase is not affected. this seems more proper.

wdyt ?

mojavelinux:
I'm not sure. That's the question I'm asking. We may need to seek input from other people to know what to do. It seems like a clear case for "quoted string".

@nicorikken any thoughts?

ldez:
I misspoke: the final render in HTML for example, two characters are replaced by a single quotation mark but the content does not change rendering.

ldez:
highlighting the whole phrase:

capture du 2016-05-19 20-15-52

nicorikken:
This feature of Asciidoctor is new to me. As the string inbetween the quotes is not affected in terms of style, highlighting the string itself seems odd to me. Or will it normalize other types of styling?
Based on a small test I get the sense something is going wrong:
```
a "`test`" b

a *"`test`"* b
```
screen shot 2016-05-20 at 08 43 42

If only the quotes are affected, maybe they need to be made bold, to show off the highlighting more clearly?

mojavelinux:
> As the string in between the quotes is not affected in terms of style, highlighting the string itself seems odd to me. 

Correct, the string in between is not affected. This syntax simply makes the quotes curved. But, like in programming, it represents a "type" (a quoted string) and we're used to seeing something that is a string highlighted as such.

I'm trying to think about what would be least distracting for the author. Maybe if we made the quotes gray or slightly transparent and left the phrase itself alone, that would be better. Or perhaps we should match, but not style by default. It's hard to say because this is new territory.

ldez:
```adoc
"`What kind of charm?`" Lazarus asked. "`An odoriferous one or a mineral one?`"
```

When I read the [documentation](http://asciidoctor.org/docs/user-manual/#curved) I understand that using context is near of [quotes](http://asciidoctor.org/docs/user-manual/#quote)

ldez:
for:

```adoc
a "`test`" b
a *"`test`"* b
```
the rendering is:

capture du 2016-05-20 23-16-30

with http://gist.asciidoctor.org

nicorikken:
Rethinking it, having a little highlighting probably wouldn't mind, as the quotes do equate to the content they encapsulate. Because the quotes have an inward direction to that content. Regarding the other inline matches, I'd like to point out a different case:
```
this "`quote with *strong* text`" is odd

this *"`strong quote`"* is just fine
```
screen shot 2016-05-21 at 10 51 24
screen shot 2016-05-21 at 10 51 16
I believe it would be better to have a new default highlight on the quoted string, which is still overridden by the other inline styles.
What do you think?

ldez:
Same case with emphasis and strong:
 
```adoc
the *foo _bar_* is here.

the _foo **bar**_ is here.
```
capture du 2016-05-21 11-01-11

nicorikken:
Point taken. This relates to the way in which inline nesting is handled in general. Then that can be directed to a separate issue, and this PR can be merged. 👍

ldez:
I opened the issue #131

PR #132, opened 2016-05-21
feat: inline assignement

ldez:
## Description

Highlighting inline attribute definition like others attribute definitions.

## Syntax example

```adoc
{set:attrname!:value}

{counter:pcount:1}
{counter2:pcount:1}

{set:foo:bar}
{set:name!}
```

## Screenshots

capture du 2016-05-21 11-55-54


mojavelinux:
Oooh, I like!

nicorikken:
Nice and clean implementation with fallback to default. I consider this ready for merge 👍

Issue #44, opened 2016-03-21
Awesome UX for "backend" attribute should be extended for other attributes


jaredmorgs:
![screen shot 2016-03-21 at 3 40 46 pm](https://cloud.githubusercontent.com/assets/952793/13911416/c272ffa0-ef7b-11e5-9026-b524f4a400f4.png)

See how the Backend attribute has microcopy help provided. 
I came across this by accident when mistyping something that the plug-in recognised.

How many other known attributes have this contextual help? 
If it's not many, we should totally extend this as much as we can. It is a great UX and reminder of what attributes do. 

What are the constraints of this microcopy? Char limits?

mojavelinux:
Now that we have identified all the attributes, we have the foundation to provide these descriptions. This may even lead us to storing them in a format that can be easily consume, which we can then use to create the attribute tables in the user manual. Now that would be *sweet*.

http://asciidoctor.org/docs/user-manual/#attribute-catalog

anthonny:
Completion is provided by the preview plugin:

https://github.com/asciidoctor/atom-asciidoc-preview/blob/master/completions.json

nicorikken:
Indeed this completion is provided by *asciidoc-preview* rather then this *language-asciidoc* package. Having completion included in a source-code oriented package would make a lot of sense. The current implementation seems to be based on an old API version, according to [the autocomplete-plus wiki](https://github.com/atom/autocomplete-plus/wiki/Autocomplete-Providers), so maybe this feature can be expanded and updated.

anthonny:
Fixed with this https://github.com/asciidoctor/atom-asciidoc-preview/issues/54 i suppose that the wiki is not up-to-date, or maybe the API has changed again

mojavelinux:
Thanks @anthonny!

We definitely need to figure out the right way to divide up the behaviors between plugins. One way to look at it is that the language plugin deals with the editor pane and the preview plugin deals with the preview pane. If putting autocomplete in the language plugin is too much, then I think we need a third plugin, atom-asciidoc-assist. I don't think autocomplete should be in the preview plugin in the future.

ldez:
Since 1.0.0 of [asciidoc-preview](https://github.com/asciidoctor/atom-asciidoc-preview), the autocompletions have been migrated to a new package: [autocomplete-asciidoc](https://atom.io/packages/autocomplete-asciidoc)

PR #137, opened 2016-05-22
feat: migrate autocomplete

### LINKED ISSUE #133: Migrate autocompletion to the grammar package.
Migrate autocompletion to the grammar package.

ldez:
- use CSON parser instead of `fs-plus` and `JSON.parse`
- `loadCompletions` become async
- use `lodash` instead of `underscore-plus`
- create a directory for completion file
- fix suggestion scope
- fix unset attribute suggestion

Fix #133, #44

Fix asciidoctor/atom-asciidoc-preview#130
Fix asciidoctor/atom-asciidoc-preview#131


nicorikken:
Nice work 👍 What was I thinking, expecting you to do a simple copy-and-paste 😉 
I haven't really used the auto-completion with intent, but the copy looks good and feature-complete. If I understand correctly the suggestions are now bound to attribute-reference `{...}`. Moving to CSON parser, and introducing better scoping looks good.

ldez:
Now that I have  working on this feature, I think it's probably a better idea to make a specific package for auto-complete like others languages.

wdyt?

nicorikken:
@ldez two remarks on your proposal:
1. We then need to have scope names coordinated across repo's/packages.
2. Would you consider snippets to be part of the `language` or the `autocomplete` package?

To me this ultimately seems to be a matter of preference between many small or few large packages.
What I like about introducing many small packages in this specific case, that the backing of the Asciidoctor 'brand' will help to enforce packages working together nicely. So in this case I do agree that a multi-package strategy can be pulled off.
Apart from subjective preference, introducing smaller packages eases the reuse of other packages, say from Markdown. Multiple ways to solve the same problem can be explored, and there is lower risk that imported packages will have to be restructured to fit into already existing (then larger) Asciidoctor packages. For example the image helper would have to be adjusted to fit into the language package.
@mojavelinux  Any project-wide guidance here?

mojavelinux:
I do like the idea of packages that do one thing and do it well (the Unix philosophy). Given that we're working under the same umbrella, I'm not too concerned about having to coordinate between them. What I am concerned about is that it makes life more complicated for the novice user who has to install each one individually.

Another idea for a combined autocomplete / snippets package name would be:

* atom-asciidoc-assistant

We could also use this as a root package name:

* atom-asciidoc-assistant-images
* atom-asciidoc-assistant-attributes
* atom-asciidoc-assistant-snippets (or templates)

(but maybe that's going too far)

To me, "autocomplete" is not descriptive enough, esp for writers. We know what it means because that's the term we use in the code, but to a less technical person, it doesn't mean much.

ldez:
'autocomplete' is the convention for Atom package.

I can create a system to auto-download packages.

ldez:
I thought about and `atom-asciidoc-assistant` is a good name :+1:

nicorikken:
Autocomplete is indeed adhering to the current Atom universe. I prefer that option for the specific auto-completion options, with a very explanatory one-liner description of course. But I can see having a namespace for a generic package.
In terms of implementation, maybe the [ide-haskell](https://atom.io/packages/ide-haskell) package can of an example. It clearly states the underlying package requirements to function properly, but no dependency resolving unfortunately.

ldez:
I agree with @nicorikken and I proposed :
- create an `atom-asciidoc-autocomplete` package with some snippets and auto-complete
- create an `atom-asciidoc-assistant` package to related all AsciiDoc tools

wdyt?

ldez:
I'm ready! I have created (in local) the `atom-autocomplete-asciidoc` package.

@mojavelinux @nicorikken Tell me what I should do?

mojavelinux:
@ldez You should now be able to move a repository into the Asciidoctor organization. I'll then add it to the atom-package-maintainers team and that will grant you admin privileges. As admin, you should also be able to manage the CI job. (I'm looking forward to the day when the same is true for AppVeyor jobs).

mojavelinux:
I wonder if the preview package should be renamed to atom-preview-asciidoc (or the autocomplete should be atom-asciidoc-autocomplete). We should probably decide on naming standards so that we end up with consistent package names.

ldez:
The naming standards already exists:
- https://atom.io/packages/search?q=preview
- https://atom.io/packages/search?utf8=%E2%9C%93&q=autocomplete

Create a custom naming standards seems to not be necessary.

mojavelinux:
I would say that a standard exists from those results. What I see is a lack of a standard. Specifically, I'm talking about the ordering (not the terminology). When does asciidoc come first and when does it come second. We should decide on a pattern and try to stick with it as we grow the number of packages.

ldez:
We can use packages provided by Atom's team as standard?
- https://atom.io/packages/autocomplete-css
- https://atom.io/packages/markdown-preview

I understand the ordering problem (if you see my initial message the ordering of the package name  is different). :wink: 

Rename a package (`asciidoc-preview`) implies that users can have the same package 2 times. :-1: 

I think we can defined naming standards for the next packages.
I propose `asciidoc-xxxxx`

wdyt?

mojavelinux:
When the package plugs into another framework, then I think we should use the following pattern:

atom-%framework%-asciidoc

* atom-language-asciidoc
* atom-autocomplete-asciidoc

When the package is adding new, original behavior, then we should use:

atom-asciidoc-%behavior%

* atom-asciidoc-preview
* atom-asciidoc-image-helper

I know that we end up with two different orderings, but at least it's for a well-specified reason.

ldez:
I transfert my repository but I don't have admin access...
- https://github.com/asciidoctor/atom-autocomplete-asciidoc

@mojavelinux can you help me?

mojavelinux:
Here I come to save the day!!

giphy

mojavelinux:
All set!

PR #138, opened 2016-05-22
feat: url and attribute reference within list definition

ldez:
## Description

url and attribute reference within list definition.

## Syntax example

```adoc
{uri-asciidoctorjs-npm}[asciidoctor.js]:: the main npm package for Asciidoctor.js
{uri-grunt-asciidoctor-npm}[grunt-asciidoctor]:: an npm package for processing AsciiDoc source files in your {uri-gruntjs}[Grunt] project

{foo_term}:: {foo_def}
```

## Screenshots

capture du 2016-05-22 17-45-33

Your screenshot.

### DIFF
--- a/grammars/language-asciidoc.cson
+++ b/grammars/language-asciidoc.cson
@@ -460,6 +460,15 @@ repository:
         name: "markup.heading.list.asciidoc"
         match: "^\\p{Blank}*(.*?\\S)(:{2,4}|;;)($|\\p{Blank}+)"
         captures:
+          "1":
+            patterns: [
+              {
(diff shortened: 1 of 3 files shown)


mojavelinux:
Nice!

Technically, any text formatting and inline macros are allowed in this position. Perhaps delegate to a group of inline patterns instead?

ldez:
We must review inline group before doing this, because some patterns like `image::`, `include::`, etc are not allowed in this element.

nicorikken:
Very nice, looks good to me

PR #143, opened 2016-05-24
refactor: remove snippets

ldez:
Remove snippets beginning with a letter.

Those snippets are now in [autocomplete-asciidoc](https://atom.io/packages/autocomplete-asciidoc)


mojavelinux:
Why not move all the snippets to the new package?

ldez:
I did not know what was best.
I can move all snippets if you think it's a good idea.

mojavelinux:
I would say that they either belong together, or we should have the snippets in a separate atom-asciidoc-snippets package. I don't think it makes sense to have the snippets split across two packages.

ldez:
I agree

ldez:
I have removed the whole file.

PR #149, opened 2016-05-30
feat: fix wrong end block

### LINKED ISSUE #148: Open block can break highlighting.
## Description

I am also experiencing an open block breaking highlighting. the `--` in the example ends the highlighting.

## Screenshots

asciidoc-language-bug

## Syntax example

Your syntax example.

```
[[zip_rollback_management_cli]]
.Rolling Back a Patch Using the Management CLI
. From the management CLI, use the `patch history` command to find the ID of the patch that you want to roll back.
+
--
NOTE: If you are using a managed domain, you must add the `--host=_HOSTNAME_` argument to the commands in this procedure to specify the {ProductShortName} host.
--

. Roll back the patch with the appropriate patch ID from the previous step.
+
[options="nowrap"]
----
patch rollback --patch-id=PATCH_ID --reset-configuration=TRUE
----
+
The patch tool will warn if there are any conflicts in attempting to roll back the patch. If there are conflicts, enter `patch --help` for the available arguments to re-run the command with an argument specifying how to resolve the conflicts.

.  Restart the {ProductShortName} server for the patch roll back to take effect:
+
[options="nowrap"]
----
shutdown --restart=true
----
```

ldez:
Fix wrong end block

Fix #148

### DIFF
--- a/grammars/language-asciidoc.cson
+++ b/grammars/language-asciidoc.cson
@@ -1155,7 +1155,7 @@ repository:
             include: "#inlines"
           }
         ]
-        end: "\\1"
+        end: "^(\\1)$"
       }
     ]
   "front-matter-block":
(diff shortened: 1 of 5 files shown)


mojavelinux:
Although it doesn't matter much, perhaps you should put the ^ and $ outside the capture. From this change, a gather those symbols doesn't carry over to the back reference.

ldez:
Fixed!

PR #156, opened 2016-06-08
feat: literal paragraph

### LINKED ISSUE #153: Match literal paragraphs
## Description

The grammar should match literal (i.e., indented) paragraphs. A literal paragraph is a paragraph that begins with one or more spaces. We should only look for spaces at the start of the paragraph, not subsequent lines within the paragraph.

## Syntax example

```asciidoc
Type the following command:

 $ gem install asciidoctor

Asciidoctor should now be installed on your machine.
```

The middle line should appear as a literal block. In other words, it's roughly the equivalent of:

```asciidoc
Type the following command:

....
$ gem install asciidoctor
....

Asciidoctor should now be installed on your machine.
```

We might even want to detect literal paragraphs that begin with `$` or `#` followed by a space and highlight them as shell commands. That's just a bonus feature as it's not technically supported in AsciiDoc (yet).

ldez:
## Description

Support literal paragraph.

## Syntax example

```adoc
Type the following command:

 $ gem install asciidoctor

Asciidoctor should now be installed on your machine.

 $ gem install asciidoctor
$ gem install asciidoctor
$ gem install asciidoctor

foo
```

## Screenshots

capture du 2016-06-08 21-04-39

Fix #153

### DIFF
--- a/grammars/language-asciidoc.cson
+++ b/grammars/language-asciidoc.cson
@@ -37,6 +37,9 @@ patterns: [
     include: "#list"
   }
   {
+    include: "#literal-paragraph"
+  }
+  {
     include: "#inlines"
   }
   {
@@ -1219,6 +1222,14 @@ repository:
         name: "markup.block.literal.asciidoc"
         begin: "^(\\.{4,})$"
         end: "^(\\1)$"
+      }
+    ]
+  "literal-paragraph":
+    patterns: [
+      {
+        name: "markup.block.literal.asciidoc"
+        begin: "^\\p{Blank}+(.*)$"
+        end: "^\\p{Blank}*$"
       }
     ]
   "open-block":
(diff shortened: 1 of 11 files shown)


ldez:
Regression with list item:

```adoc
* Asciidoctor strips XML entities from the section title before
  generating the id (makes for cleaner section ids)
```

grammars/language-asciidoc.cson
```diff
@@ -1221,6 +1224,14 @@ repository:
         end: "^(\\1)$"
       }
     ]
+  "literal-paragraph":
+    patterns: [
+      {
+        name: "markup.block.literal.asciidoc"
+        begin: "^\\p{Blank}+(.*)$"
```
mojavelinux: I think this should be `.+` (instead of `.*`). There needs to be at least one non-whitespace character in the line.

ldez: currently in WIP

nicorikken: Were you thinking of more changes other than `.+` to `.*` ? Otherwise it seems completing this enhancement would be relatively easy, as you already did most of the work.

mojavelinux:
Did you merge this? I can't tell from the activity history.

PR #162, opened 2016-09-17
Add grammar for general inline macro

ProgramFan:
Add general inline macro grammar so that custom inline macros can be highlighted. 

For example: `cite:[XXX]` can be highlighted as in the following screenshot:


mojavelinux:
This is important for promoting the idea that extensions are a natural part of the AsciiDoc language.

Is the intent to style custom macros differently than built-in macros? I can see the logic to that. If not, all inline macros (including links) follow the same pattern, so it might be more efficient just to capture them all at once.

ProgramFan:
Sorry for the mess of commit history. But the diff shall be clean. Would someone review it? Thanks.

ldez:
Please rebase you branch. Thanks.

ProgramFan:
Done. @ldez

ldez:
I need to fix #170 before review it.

ProgramFan:
OK.

ldez:
Can you rebase once again, thanks.

ProgramFan:
Done.

PR #185, opened 2017-09-27
Do not attempt to match newline characters

50Wliu:
## Description

This PR removes the `[\\r\\n]+` requirement present in some end patterns.  This regex wasn't really doing anything important in the first place (all it was contributing was a zero-width match at the end of the line).  In addition, due to atom/first-mate#100, newlines are no longer added to the last line of the file, meaning that rules containing `[\\r\\n]+` would fail to match on the last line.  Therefore, this change is simply a backwards-compatible change to ensure that language-asciidoc continues working as before on Atom 1.22+.

I hope I did the whole grammar generation correctly!

/cc @Ingramz

Refs atom/atom#15729

### DIFF
--- a/grammars/language-asciidoc.cson
+++ b/grammars/language-asciidoc.cson
@@ -1136,7 +1136,7 @@ repository:
             end: "(?<=\\1)"
           }
         ]
-        end: "((?<=--|====)[\\r\\n]+$|^\\p{Blank}*$)"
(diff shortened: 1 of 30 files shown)


ldez:
@50Wliu Thanks 👍 

The grammar generation is right.

I don't see any regressions.

